PR #28, opened 2026-03-17
Fix SpiralDataProvider empty URL fallthrough to env var

pcalnon:
## Summary
- **Bug**: `SpiralDataProvider.__init__` used Python's `or` operator to resolve `juniper_data_url`, which treats `""` as falsy — causing an explicitly passed empty string to silently fall through to the `JUNIPER_DATA_URL` environment variable
- **Fix**: Replace `or` with `is not None` check and `.strip()` normalization, so callers can explicitly pass `""` to disable JuniperData integration
- **Tests**: Added 7 new unit tests covering empty string, whitespace-only, `None` fallback to env var, and error paths for unconfigured URLs

## Test plan
- [x] `pytest unit/test_spiral_data_provider.py -v` — 32/32 passed
- [x] Full test suite via `scripts/run_tests.bash` — all passed (95% coverage)


pcalnon (commented):
approved

PR #69, opened 2026-04-04
fix: restore early stopping and raise on unrecognized activation (CR-040, CR-043, CR-046)

pcalnon:
## Summary

- **CR-040**: Restore `patience_counter` and `best_value_loss` propagation between `grow_network` iterations — early stopping patience was completely non-functional because validation results were never read back into loop variables (2-line fix)
- **CR-043**: Uncomment training-loss early stopping path when no validation data is provided — the `else`-block in `validate_training` was entirely commented out, making early stopping impossible without validation data
- **CR-046**: Raise `ValueError` instead of silently falling back to ReLU when `ActivationWithDerivative` encounters an unrecognized activation name during deserialization

## Impact

These three findings from the comprehensive code review are the highest-priority core ML correctness fixes:
- **CR-040 (S1)** is the single highest-impact fix — patience-based early stopping was completely broken
- **CR-043 (S2)** completes the early stopping fix by enabling training-loss-based stopping when no validation data is available
- **CR-046 (S2)** prevents silent behavior changes from corrupting training results during multiprocessing

## Test plan

- [x] Updated `test_unknown_activation_fallback` → `test_unknown_activation_raises_value_error` to assert `ValueError` on unrecognized activation deserialization
- [x] Full test suite passes (2701 passed, 234 skipped, 0 failed)
- [ ] Verify early stopping triggers correctly in integration training run with patience < max_iterations

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -3746,16 +3746,9 @@ class CascadeCorrelationNetwork:
                 traceback.print_exc()
                 raise TrainingError from e
 
-            # # Update loop state from validation results (critical for early stopping convergence)
-            # patience_counter = validate_training_results.patience_counter
-            # best_value_loss = validate_training_results.best_value_loss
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #73, opened 2026-04-04
fix: auto-reset state machine from terminal states on start (CR-007)

pcalnon:
## Summary

- **Option A**: `_handle_start` now auto-resets from FAILED/COMPLETED terminal states before proceeding with the start transition, preventing permanent state machine desynchronization
- **Option C**: Removed duplicate error handler in `_run_training` that redundantly attempted `Command.STOP` from FAILED state (producing spurious "Invalid transition" warnings) — exception handling is performed by `monitored_fit()`

## Impact

After training fails or completes, the state machine previously entered an irrecoverable terminal state. Calling `start_training` again would bypass the guard (it only checks `is_started()`), submit new work, but `monitored_fit`'s `Command.START` would silently fail. The state machine would report stale data while training actually executed.

## Test plan

- [x] Added `test_start_auto_resets_from_failed` — verifies start succeeds from FAILED state
- [x] Added `test_start_auto_resets_from_completed` — verifies start succeeds from COMPLETED state
- [x] Updated `test_run_training_propagates_exception` to reflect new contract
- [x] Full test suite passes (2703 passed, 234 skipped, 0 failed)

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -531,23 +531,13 @@ class TrainingLifecycleManager:
         return {"status": "training_started", "timestamp": time.time()}
 
     def _run_training(self, x, y, x_val, y_val, **kwargs) -> None:
-        """Execute training in background thread."""
-        try:
-            self.network.fit(x, y, x_val=x_val, y_val=y_val, **kwargs)
-        except Exception as e:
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #95, opened 2026-04-05
fix: snapshot round-trip fidelity for activation functions (CR-048, CR-044)

pcalnon:
## Summary

- **CR-048**: Fix activation name extraction in `_save_hidden_units` — use `_activation_name` attribute instead of `__name__` which doesn't exist on `ActivationWithDerivative` wrappers. Per-unit activation identity was silently lost on save when units had heterogeneous activations.
- **CR-044**: Wrap deserialized activation functions in `ActivationWithDerivative` during HDF5 load, matching runtime-created units. Previously loaded units got bare `torch.nn` modules while runtime units got wrapped versions.
- Also fixes duplicate `max_iterations` parameter in `fit()` signature introduced during prior merge.

## Test plan

- [x] Full unit test suite passes (0 failures)
- [ ] Round-trip test: save network with heterogeneous per-unit activations → load → verify `_activation_name` preserved and derivatives work

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -1372,7 +1372,6 @@ class CascadeCorrelationNetwork:
         epochs: int = None,
         max_iterations: int = None,
         early_stopping: bool = True,
-        max_iterations: int = None,
     ) -> Dict[str, List]:
         """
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #97, opened 2026-04-05
fix: whitelist training start params and bound inline dataset arrays (CR-023, CR-029)

pcalnon:
## Summary

- **CR-023**: Add parameter whitelist in `start_training` route — only allowed keys from `body.params` are forwarded as `**kwargs`. Unrecognized keys are logged and dropped, preventing information disclosure from unexpected kwargs leaking internal function signatures.
- **CR-029**: Add `max_length=100000` to `InlineDataset` array fields. Inline data is for small ad-hoc datasets; large datasets should use the juniper-data service.

## Test plan

- [x] 610 API unit tests pass
- [x] Also fixes duplicate `max_iterations` parameter in `fit()` from prior merge
- [ ] Test with request containing unrecognized params key → verify it's logged and dropped

### DIFF
--- a/src/api/models/training.py
+++ b/src/api/models/training.py
@@ -15,12 +15,16 @@ class DatasetSource(BaseModel):
 
 
 class InlineDataset(BaseModel):
-    """Inline dataset provided directly in the request body."""
+    """Inline dataset provided directly in the request body.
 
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #99, opened 2026-04-05
fix: add worker_id format validation in registration (CR-026)

pcalnon:
## Summary

- Add format validation for `worker_id` in `WorkerProtocol.validate_register()`: alphanumeric + hyphens + underscores, 1-64 chars, must start with alphanumeric
- Previously workers could supply arbitrary strings (including empty, special chars, excessively long) as their identity with no server-side validation

## Test plan

- [x] 95 worker protocol/stream/registry tests pass
- [ ] Registration tokens for reconnection (deferred per review recommendation)

### DIFF
--- a/src/api/workers/protocol.py
+++ b/src/api/workers/protocol.py
@@ -19,6 +19,7 @@ Message Types:
     error          — Either direction: error notification
 """
 
+import re
 import struct
 import time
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #102, opened 2026-04-05
fix: chunked body limit and async WebSocket lock (CR-024, CR-025)

pcalnon:
## Summary

- **CR-024 (S2, security)**: `RequestBodyLimitMiddleware` now reads and measures the body for POST/PUT/PATCH requests with chunked transfer encoding (no Content-Length header). Previously the 10MB limit was entirely bypassable.
- **CR-025 (S2, concurrency)**: Add `asyncio.Lock` to `WebSocketManager` for `connect()` and `disconnect()` — ensures `_active_connections` and `_connection_meta` mutations are atomic regardless of event loop implementation.

## Test plan

- [x] 610 API unit tests pass


pcalnon (commented):
approved

PR #104, opened 2026-04-05
perf: lazy logging in candidate training hot path (CR-062)

pcalnon:
## Summary

Reduce logging overhead in `candidate_unit.train_detailed()` inner loop from ~18 f-string tensor interpolations per epoch to ~4 guarded lazy calls:

- Hoist constant tensor metadata (residual_error shape/dtype/dims) before the loop
- Guard DEBUG/TRACE logging with `isEnabledFor()` to skip f-string evaluation when level is disabled
- Convert remaining per-epoch logs to `%s` lazy formatting
- Remove 8 redundant debug logs in the display progress block (logged object repr every epoch)

## Impact

For 16 candidates * 200 epochs = 3,200 iterations: eliminates ~57,600 unnecessary string interpolations per training round at production log levels (INFO). The conftest `_NoOpLogger` was created specifically to work around this overhead (~0.9s per `fit()` call).

## Test plan

- [x] Full unit test suite passes (exit code 0)

### DIFF
--- a/src/candidate_unit/candidate_unit.py
+++ b/src/candidate_unit/candidate_unit.py
@@ -554,29 +554,28 @@ class CandidateUnit:
         actual_epochs_completed = 0
         self.logger.debug(f"CandidateUnit: train: Early stopping enabled: {self.early_stopping}, Patience: {self.patience}")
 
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #105, opened 2026-04-05
fix: thread-safe audit counters and stale-worker check (CR-027, CR-028)

pcalnon:
## Summary
- **CR-027**: Add `threading.Lock` to `AuditLogger._counter` and `WorkerMetrics._workers` to prevent race conditions on compound read-modify-write operations
- **CR-028**: Re-check worker staleness before deregistration in `_check_stale_workers` to close TOCTOU gap where a worker could heartbeat between snapshot and removal

## Test plan
- [x] All 312 unit tests pass
- [ ] Verify lock contention doesn't impact worker throughput under load

### DIFF
--- a/src/api/workers/audit.py
+++ b/src/api/workers/audit.py
@@ -16,6 +16,7 @@ import json
 import logging
 import time
 from enum import StrEnum
+from threading import Lock
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #106, opened 2026-04-05
fix: mirror _roll_sequence_number capped-loop fix in network class (CR-041)

pcalnon:
## Summary
- Mirror the CASCOR-P1-008 fix from CandidateUnit in `CascadeCorrelationNetwork._roll_sequence_number`: replace list comprehension with capped for-loop to prevent unbounded memory allocation
- Increase `_PROJECT_SEQUENCE_MAX_VALUE` from 10 to 100 for consistency across both classes

## Test plan
- [x] All unit tests pass
- [ ] Verify sequence number generation still produces expected randomization


pcalnon (commented):
approved

PR #110, opened 2026-04-05
fix: test quality improvements and documentation (CR-061, CR-063, CR-073, CR-074, CR-076)

pcalnon:
## Summary
- **CR-061**: Document intentional optimizer reset in `train_output_layer` (parameter space changes when hidden units are added)
- **CR-063**: Store metadata only (shape, correlation, unit index) in unit history instead of weight copies — reduces memory duplication. Backward-compatible deserialization for legacy snapshots.
- **CR-073**: Add smoke test running `fit()` with real logger at WARNING level to catch logging attribute errors
- **CR-074**: Rewrite `test_grow_network_adds_hidden_unit` to actually call `grow_network()` with minimal parameters
- **CR-076**: Remove 6 redundant per-fixture `torch.manual_seed(42)` calls; `reset_random_seeds` autouse fixture is single source of truth

## Test plan
- [x] All modified tests pass (28 passed, 2 skipped)
- [x] All 102 snapshot serializer tests pass (backward compatibility verified)
- [x] Full unit suite passes (1 pre-existing failure in `test_api_observability.py`)

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -1472,7 +1472,7 @@ class CascadeCorrelationNetwork:
             x_val=x_val,
             y_val=y_val,
         )
-        self.history["hidden_units_added"].append({"correlation": 0.0, "weights": [], "bias": []})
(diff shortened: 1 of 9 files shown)


pcalnon (commented):
approved

PR #112, opened 2026-04-06
perf: extract _compute_hidden_outputs and hoist above epoch loop (CR-060, CR-050)

pcalnon:
## Summary
- Extract shared `_compute_hidden_outputs()` using OPT-1 pre-allocated buffer pattern, replacing 5 duplicate implementations across `forward()`, `_prepare_candidate_input()`, `train_output_layer()`, `add_unit()`, and `add_units_as_layer()`
- **Key performance win**: Hoist hidden output computation above the epoch loop in `train_output_layer()` — hidden weights are frozen during output training, so outputs are constant across epochs. Eliminates N×E redundant forward passes (where N = hidden units, E = epochs)
- Net change: 46 insertions, 67 deletions (consolidation reduces code)

## Test plan
- [x] All unit tests pass
- [ ] Benchmark training throughput before/after to quantify speedup
- [ ] Verify numerical equivalence with snapshot round-trip test

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -1479,6 +1479,36 @@ class CascadeCorrelationNetwork:
         return self.history
 
     #################################################################################################################################################################################################
+    # Private helper: compute hidden unit outputs using pre-allocated buffer (OPT-1)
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #113, opened 2026-04-06
fix: replace time.sleep with polling helper in integration tests (CR-070)

pcalnon:
## Summary
- Add `wait_for_state()` polling helper that actively polls training status until the expected state is reached, with configurable timeout and poll interval
- Replace all 11 fixed `time.sleep()` calls across 6 test methods with deterministic state-based polling
- Eliminates flaky timing dependencies under CI load while maintaining test correctness

## Test plan
- [x] All 7 integration tests pass
- [x] Unit suite passes (1 pre-existing failure in `test_api_observability.py`)
- [ ] Verify under CI load that polling timeouts are sufficient

### DIFF
--- a/src/tests/integration/api/test_api_full_lifecycle.py
+++ b/src/tests/integration/api/test_api_full_lifecycle.py
@@ -13,6 +13,36 @@ from api.app import create_app
 from api.settings import Settings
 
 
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #120, opened 2026-04-09
fix: Phase 2 security hardening — CR-024, CR-025, CR-026 + CR-023 test

pcalnon:
## Summary

Phase 2 of the Canopy-Cascor interface review roadmap — security hardening. Cascor-only; no canopy coordination required. No dependency on Phase 1 PRs (the roadmap explicitly lists these phases as parallelizable).

- **CR-023** (S2) — verified already resolved on main; added regression test closing the coverage gap
- **CR-024** (S2) — **FIXED**: chunked encoding bypass via streaming body reader with per-chunk cap
- **CR-025** (S2) — **FIXED**: `WebSocketManager.close_all()` now takes snapshot under lock (connect/disconnect were already locked)
- **CR-026** (S1) — **FIXED**: server now generates UUID-based worker IDs; client-proposed name kept as untrusted display label only

## Approach

Followed the same verification-first approach as Phase 1. Read each issue's current state in the codebase before touching anything:

| Issue | Pre-fix state | Action |
|---|---|---|
| CR-023 | Whitelist at `routes/training.py:36-42` already filters unknown params with warning log | Added regression test `test_start_training_filters_unwhitelisted_params` |
| CR-024 | `await request.body()` allocated full body before length check — memory exhaustion via chunked stream | Stream-read with cumulative byte cap, early-abort at 413; Content-Length retained as fast-path only |
| CR-025 | Lock covered `connect()`/`disconnect()` but not `close_all()` — shutdown-vs-connect race | `close_all()` now snapshots under lock, releases lock, then closes |
| CR-026 | `worker_id = msg["worker_id"]` accepted client-supplied identity — trivial impersonation | Server generates `worker-<12 hex chars>` UUID; client name captured as `client_name` for audit only |

## New test coverage

14 new regression tests across 5 files:

- `test_api_middleware.py` — 7 tests covering CR-024 (small body, GET not affected, declared Content-Length over limit early-reject, invalid Content-Length, actual body over limit, chunked under limit, chunked over limit via generator)
- `test_websocket_manager.py` — 1 test covering CR-025 (`close_all` blocks when lock is externally held)
- `test_worker_stream.py` — 3 tests covering CR-026 (server != client id, registration_ack returns server id, impersonation via same client_name)
- `test_worker_security_integration.py` — 1 updated test (metrics lookup by server id, plus negative assertion that client name never leaks into metrics keyspace)
- `test_training_route_coverage.py` — 1 test covering CR-023 (spies on `lifecycle.start_training` to verify non-whitelisted keys never reach forwarded kwargs)

## Backwards compatibility

- **CR-026**: Wire protocol unchanged. The REGISTER payload still has a `worker_id` field (now ignored as identity, captured as `client_name`). The `registration_ack` still returns a `worker_id` field (now the server-assigned UUID). Worker clients that already consume the ack's `worker_id` are unaffected. Worker clients that hard-code their own proposed ID for subsequent out-of-band references (e.g. dashboards) will need to use the server-returned ID instead.
- **CR-024**: Transparent to well-behaved clients. Only affects clients that were sending lying Content-Length headers or using chunked streams > 10MB.
- **CR-025**: Pure internal locking change, no API surface impact.

## Test plan

- [x] `pytest tests/unit/api/` → all tests pass, exit=0
- [x] 14 new regression tests all passing
- [ ] Manual end-to-end smoke: connect a worker and confirm it receives a server-assigned ID via the ack

## Companion PR

- juniper-ml docs PR: ROADMAP and ANALYSIS updated with Phase 2 execution results (to follow)

### DIFF
--- a/src/api/middleware.py
+++ b/src/api/middleware.py
@@ -53,8 +53,18 @@ _MAX_REQUEST_BODY_BYTES = 10 * 1024 * 1024  # 10 MB
 class RequestBodyLimitMiddleware(BaseHTTPMiddleware):
     """Reject requests whose body exceeds a configurable limit.
 
-    Checks Content-Length header when present; for chunked transfer encoding
-    (no Content-Length), reads and measures the body incrementally.
+    The ``Content-Length`` header is used as an **early reject** fast path but
+    is not trusted as the sole size check (CR-024): a malicious client can
+    under-declare or omit the header and send an unbounded chunked stream.
+    For POST/PUT/PATCH requests we always stream-read the body with a
+    cumulative byte cap, aborting with HTTP 413 as soon as the cap is
+    exceeded. This prevents the classic chunked-encoding memory-exhaustion
+    bypass in which ``await request.body()`` would allocate the entire body
+    before any size check runs.
+
+    The fully-read body is cached on ``request._body`` so downstream FastAPI
+    route handlers can consume it via ``request.body()`` / ``request.json()``
+    / pydantic body parsing without triggering a second read.
     """
 
     def __init__(self, app: ASGIApp, max_bytes: int = _MAX_REQUEST_BODY_BYTES) -> None:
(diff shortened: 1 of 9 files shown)


pcalnon (commented):
approved

PR #121, opened 2026-04-10
fix: align epochs_max default to canopy's 1,000,000

pcalnon:
## Summary

- Aligns the cascor API model \`epochs_max\` default from 200 to 1,000,000 to match canopy's \`nn_max_total_epochs\` default
- Touches \`api/models/network.py:21\` (NetworkCreateRequest field default) and \`api/lifecycle/manager.py:176\` (kwargs fallback)
- Adds regression test \`test_create_network_epochs_max_default_aligned_with_canopy\`
- Resolves Phase 1 deferred item from the Canopy-Cascor interface roadmap

## Behavior change

This is an intentional behavior change for cascor API clients that don't pass \`epochs_max\` explicitly: they will now train up to 1M epochs instead of 200. Prior 200 default silently capped training at a level vastly below the canopy slider range (10 to 10,000,000, default 1,000,000), surprising users who edited other params via the canopy UI.

## Test plan

- [x] All cascor api unit tests pass (exit 0)
- [x] New regression test locks in the 1M default

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -173,7 +173,9 @@ class TrainingLifecycleManager:
                 phase="Idle",
                 learning_rate=kwargs.get("learning_rate", 0.01),
                 max_hidden_units=kwargs.get("max_hidden_units", 1000),
-                max_epochs=kwargs.get("epochs_max", 200),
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #122, opened 2026-04-10
chore: raise training limit constants per canopy requirements

pcalnon:
## Summary

Raises the cascor training-limit defaults per 2026-04-10 canopy requirements:

| Constant | Old | New |
|---|---|---|
| \`epochs_max\` | 1,000,000 (1e6) | 100,000,000,000 (1e11) |
| \`max_iterations\` | 1,000 (1e3) | 1,000,000 (1e6) |
| \`max_hidden_units\` | 1,000 (1e3) | 10,000 (1e4) |

Touches:
- \`src/cascor_constants/constants_model/constants_model.py\` (active values in the menu format)
- \`src/api/models/network.py\` (NetworkCreateRequest field defaults)
- \`src/api/lifecycle/manager.py\` (kwargs.get fallbacks)
- \`src/tests/unit/api/test_lifecycle_manager.py\` (regression test renamed and expanded)

## Behavior change

Cascor API clients that don't pass these fields explicitly will now train up to 100 billion epochs / 1 million growth iterations / 10,000 hidden units instead of the previous lower caps. The intent is that the user, not the API model, decides when to stop. Convergence-based stopping criteria still apply.

## Test plan

- [x] All cascor src/tests/unit/api tests pass (exit 0)
- [x] Regression test \`test_create_network_training_limit_defaults_aligned_with_canopy\` asserts all three new defaults

Companion PR: pcalnon/juniper-canopy with the same constants update on the canopy side.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -172,11 +172,12 @@ class TrainingLifecycleManager:
                 status="Stopped",
                 phase="Idle",
                 learning_rate=kwargs.get("learning_rate", 0.01),
-                max_hidden_units=kwargs.get("max_hidden_units", 1000),
-                # epochs_max default aligned with canopy nn_max_total_epochs (1,000,000) — see
+                max_hidden_units=kwargs.get("max_hidden_units", 10000),
+                # Training-limit defaults raised per canopy requirements (2026-04-10):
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #138, opened 2026-04-24
fix(phase-2a): data-integrity bugs BUG-CC-03, BUG-CC-11, BUG-CC-18/ROBUST-01

pcalnon:
## Summary

Implements **Track 2 / Phase 2A** of the V7 roadmap — three correctness fixes that together eliminate silent data-corruption paths in juniper-cascor.

- **BUG-CC-11 (P0)** — Parenthesize walrus in `utils._object_attributes_to_table` so `content` receives the list, not a bool.
- **BUG-CC-03 (P1)** — Replace `param or self.param or DEFAULT` chains with explicit `is not None` checks in `SpiralProblem` parameter-init methods so valid falsy overrides (`False`, `0`, `0.0`) are preserved.
- **BUG-CC-18 / ROBUST-01 (P0, Critical)** — Introduce `CandidateTrainingError` and raise it from `_execute_candidate_training` on double-failure / empty-results paths instead of silently installing zero-correlation dummy candidates. `train_candidates` now re-raises the specific error unchanged.

## Test plan

- [x] New `tests/unit/test_phase_2a_data_integrity.py` (9 regression tests) all pass.
- [x] Updated `test_utils_coverage.py`, `test_remaining_coverage_deep.py`, and `test_cascade_correlation_coverage_deep.py` now lock in post-fix behavior.
- [x] Broader unit-test runs covering affected files (`utils`, `spiral_problem`, `cascade_correlation` coverage/extended/deep, `candidate_training_manager`, `candidate_unit_coverage`, `config_and_exceptions`) pass without regressions.
- [ ] CI is the source of truth — confirm on the PR.

Notes:
- Pre-existing stale test files referencing modules removed in commit cbcf9bd (`test_spiral_check_coverage.py`, `test_remote_client_0_extended.py`) continue to collect-error — out of scope for Phase 2A; tracked separately.
- `_get_dummy_results` is retained for the defensive path in `_process_training_results`; only the two silent-corruption callsites are removed.

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -82,7 +82,7 @@ from candidate_unit.candidate_unit import CandidateTrainingResult, CandidateUnit
 #
 # Server-owned queues (live in Manager server process)
 from cascade_correlation.cascade_correlation_config.cascade_correlation_config import CascadeCorrelationConfig
-from cascade_correlation.cascade_correlation_exceptions.cascade_correlation_exceptions import ConfigurationError, TrainingError, ValidationError  # CascadeCorrelationError,; NetworkInitializationError,
+from cascade_correlation.cascade_correlation_exceptions.cascade_correlation_exceptions import CandidateTrainingError, ConfigurationError, TrainingError, ValidationError  # CascadeCorrelationError,; NetworkInitializationError,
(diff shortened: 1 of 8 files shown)


pcalnon (commented):
approved

PR #140, opened 2026-04-25
fix(phase-2c): juniper-cascor memory leaks and body-limit bypass (BUG-CC-13/14/15)

pcalnon:
## Summary

Phase 2C of the Juniper outstanding-items roadmap. Three independent fixes in `juniper-cascor`:

- **BUG-CC-13** (MEDIUM, `src/api/security.py`): `RateLimiter._counters` had no eviction mechanism, leading to unbounded memory growth (one entry per unique key/IP forever). Added `_maybe_cleanup()` mirroring the existing `ConnectionRateLimiter` pattern: runs every 100 `check()` calls, evicts buckets older than `2 * window_seconds`, and hard-caps at 10,000 entries (oldest by window_start dropped first).
- **BUG-CC-14** (LOW, `src/api/websocket/control_security.py`): `HandshakeCooldown._rejections` only pruned the *current* IP per-call, leaving stale histories for non-blocked IPs forever. Added `_maybe_full_cleanup()` that runs every 50 `record_rejection()` calls and removes IPs whose timestamps are all older than `2 * window_sec`.
- **BUG-CC-15** (HIGH, `src/api/middleware.py`): `RequestBodyLimitMiddleware` called `await request.body()` on the no-Content-Length path, allocating the full body before the size check (SEC-08 partial reopening; the docstring claimed streaming, but the implementation did not). Replaced with `async for chunk in request.stream()` plus an early 413 abort when cumulative bytes exceed `max_bytes`. Body is cached on `request._body` so downstream FastAPI handlers can still read it.

All changes follow the recommended approaches from the roadmap (notes/JUNIPER_OUTSTANDING_DEVELOPMENT_ITEMS_V7_IMPLEMENTATION_ROADMAP.md sections 2053+, 2132+, 2201+).

## Test coverage added

- `src/tests/unit/api/test_api_security.py::TestRateLimiterBugCC13Cleanup` (4 tests)
  - Expired entries pruned after CLEANUP_INTERVAL
  - Hard cap (`_MAX_ENTRIES`) prevents unbounded growth even with all-fresh keys
  - Fresh entries within active window are not evicted
  - `reset()` clears the cleanup tick counter
- `src/tests/unit/api/test_control_security.py::TestHandshakeCooldownBugCC14Cleanup` (3 tests)
  - Stale non-blocked IPs pruned after CLEANUP_EVERY_N rejections
  - Blocked IPs cleared from `_rejections` (existing behavior preserved)
  - Fresh non-blocked IPs preserved across cleanup
- `src/tests/unit/api/test_api_middleware.py` (2 tests)
  - `test_bug_cc_15_chunked_body_aborts_before_full_buffer` — counts bytes consumed via mocked `request.stream()` and asserts the middleware aborts within one chunk past the limit
  - `test_bug_cc_15_body_cached_for_downstream_handlers` — verifies cached body remains readable by downstream handlers via TestClient

Total: 9 new tests. All 75 existing+new tests in `test_api_security.py`, `test_control_security.py`, and `test_api_middleware.py` pass. Full `src/tests/unit/api/` suite passes (rc=0).

## Test plan

- [ ] CI pre-commit hooks green
- [ ] CI unit tests green
- [ ] CI integration tests green
- [ ] Manual review of memory bounds (`_MAX_ENTRIES=10000`, `CLEANUP_INTERVAL=100`, `CLEANUP_EVERY_N=50`)

## Notes

- No backwards-compat shims; behavior is purely additive (eviction of stale entries / earlier 413 on oversized streamed bodies).
- One inline comment per fix site tagged with the bug ID for traceability.
- No deviations from the roadmap's recommended approach.

### DIFF
--- a/src/api/middleware.py
+++ b/src/api/middleware.py
@@ -88,9 +88,16 @@ class RequestBodyLimitMiddleware(BaseHTTPMiddleware):
             if declared_length > self._max_bytes:
                 return JSONResponse(status_code=_PROJECT_API_HTTP_413_PAYLOAD_TOO_LARGE, content={"detail": "Request body too large"})
         if content_length is None and request.method in ("POST", "PUT", "PATCH"):
-            body = await request.body()
-            if len(body) > self._max_bytes:
-                return JSONResponse(status_code=_PROJECT_API_HTTP_413_PAYLOAD_TOO_LARGE, content={"detail": "Request body too large"})
+            # BUG-CC-15: stream-read with early abort to avoid buffering full body before size check.
+            chunks: list[bytes] = []
+            size = 0
+            async for chunk in request.stream():
+                size += len(chunk)
+                if size > self._max_bytes:
+                    return JSONResponse(status_code=_PROJECT_API_HTTP_413_PAYLOAD_TOO_LARGE, content={"detail": "Request body too large"})
+                chunks.append(chunk)
+            # Cache body for downstream handlers (Starlette convention).
+            request._body = b"".join(chunks)
         return await call_next(request)
 
 
--- a/src/api/security.py
+++ b/src/api/security.py
@@ -1,6 +1,7 @@
 """API security: authentication and rate limiting middleware."""
 
 import hmac
+import logging
 import time
 from collections import defaultdict
 from threading import Lock
@@ -9,6 +10,8 @@ from fastapi import HTTPException, Request, status
 from fastapi.security import APIKeyHeader
 
 from .settings import get_settings
+
+logger = logging.getLogger(__name__)
 
 api_key_header = APIKeyHeader(name="X-API-Key", auto_error=False)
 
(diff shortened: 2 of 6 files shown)


pcalnon:
## CI status note

All four GitHub Actions workflow runs (`ci.yml`, `security-scan.yml`, `publish.yml`, `lockfile-update.yml`) failed at startup (0s, no jobs executed). Investigation:

- `gh workflow run ci.yml --ref fix/phase-2c-memory-leaks-body-limit` reports YAML syntax errors at `.github/workflows/ci.yml` lines 121, 456, 518: `A sequence was not expected` (e.g. `python-version: ["3.12", "3.13", "3.14"]` passed to `actions/setup-python` `with:`, which expects a string).
- The same workflow failures are present on `main` for commit 667d148 (Phase 1C) and prior commits — this is a pre-existing repo-wide CI infrastructure issue, NOT introduced by this PR.
- `git diff main HEAD -- .github/workflows/` is empty.

Per the Phase 2C scope constraint ("Do not bundle in unrelated cleanup"), the workflow file fix is intentionally out of scope for this PR and should be addressed separately.

### Local verification

All checks pass locally on this branch:
- `pre-commit run --files <changed files>` — all hooks Passed
- `pytest src/tests/unit/api/` — rc=0 (full API unit suite)
- 9 new regression tests added (4 BUG-CC-13, 3 BUG-CC-14, 2 BUG-CC-15) all passing

PR #142, opened 2026-04-26
fix: lock broadcast throttle and metrics extraction (CONC-02/CONC-03, Track 3 Phase 3B)

pcalnon:
## Summary

Track 3 / Phase 3B — two related concurrency fixes in `src/api/lifecycle/manager.py`.

### CONC-02 / BUG-CC-16 — `_last_state_broadcast_time` unprotected R/W

`_broadcast_training_state` formerly read and wrote `_last_state_broadcast_time` outside any lock. Two callers (training thread, monitor thread, control endpoint) could both observe `now - last >= interval`, both pass the throttle, and both broadcast — defeating the GAP-WS-21 coalescer.

Fix: allocate `self._broadcast_lock = threading.Lock()` in `__init__`, initialize `_last_state_broadcast_time = 0.0` (eliminating the earlier `hasattr` gate), and hold the lock across the throttle check + timestamp update so only one caller can win each window. Terminal transitions and `force=True` still bypass the throttle but still update the timestamp under the lock for consistency.

### CONC-03 / BUG-CC-17 — `_extract_and_record_metrics` split-lock

The previous implementation held `self._metrics_lock` only briefly to snapshot `network.history` + read the high-water-mark, released it across the per-entry `on_epoch_end` calls, and re-acquired it to advance the high-water-mark. Two concurrent callers could both observe the same `last_emitted = 0`, both emit identical epoch-1..N entries, and only then race on the write — so each epoch was reported to `TrainingMonitor` twice.

Fix: hold `self._metrics_lock` across the entire read-process-write cycle: snapshot, per-entry `on_epoch_end` calls, and high-water-mark advance. The idempotent `training_state.update_state` call is left outside the lock to bound the critical section.

### Tests

Adds `src/tests/unit/api/test_lifecycle_concurrency.py` with 4 tests:

- **`TestBroadcastThrottleRace`** — patches `time.monotonic` to return identical ticks under a barrier so 16 concurrent broadcasters all observe the same throttle window. Without the lock multiple `broadcast_from_thread` calls fire; with the lock exactly one wins.
- **`TestExtractAndRecordMetricsRace`** — patches `TrainingMonitor.on_epoch_end` to sleep ~5 ms so the per-entry loop becomes observable; without the single-lock scope, 8 concurrent extractors each emit all `n_epochs` entries; with the fix every epoch is emitted exactly once. Plus an idempotency check that a second call with no new history doesn't re-emit anything.

## Test plan

- [x] Static lint pass: `black --check`, `flake8`, `isort --check` against the changed files — all clean.
- [x] Static type check via `python -m py_compile` against `src/api/lifecycle/manager.py` and the new test file.
- [ ] Local pytest run is **blocked by a pre-existing torch C-extension import failure** in the JuniperCascor conda env (`torch._C` directory shadowing `torch._C.so` under Python 3.14 free-threading) — same root cause documented in the auto-memory; affects every cascor test that transitively imports `api.lifecycle.manager`. Trust CI / admin-merge per the `juniper-cascor CI Workflows Broken` workflow note.
- [x] CONC-02 and CONC-03 critical sections preserved in single `with self._lock:` blocks with explanatory comments referencing the issue IDs.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -43,6 +43,13 @@ class TrainingLifecycleManager:
         self._training_lock = threading.Lock()
         self._metrics_lock = threading.Lock()
         self._topology_lock = threading.Lock()
+        # CONC-02 / BUG-CC-16 (Phase 3B): guard the broadcast throttle so two
+        # callers cannot both pass the (now - last) < interval check and emit
+        # duplicate state messages. _last_state_broadcast_time is initialized
+        # here so the read in _broadcast_training_state never has to gate on
+        # hasattr() (which itself was part of the original race window).
+        self._broadcast_lock = threading.Lock()
+        self._last_state_broadcast_time: float = 0.0
         self._executor: Optional[ThreadPoolExecutor] = None
         self._training_future: Optional[Future] = None
         self._stop_requested = threading.Event()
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #144, opened 2026-04-27
fix: track startup task references in lifespan (CONC-09, Track 3 Phase 3C)

pcalnon:
## Summary

Track 3 / Phase 3C — **CONC-09**: store fire-and-forget auto-start tasks on `app.state.startup_tasks` and surface their exceptions.

`src/api/app.py::lifespan` previously created the `auto_start_training` and `auto_start_canopy` tasks with `asyncio.create_task(...)` and discarded the return value. Per the asyncio docs that's the canonical fire-and-forget anti-pattern: the loop is free to GC the task mid-flight, and any exception raised inside it is silently swallowed (only later surfaced as the cryptic "Task exception was never retrieved" warning at GC time, with no traceback).

### Changes

- New `_log_startup_task_exception(task)` done-callback that logs non-cancellation exceptions at error level with `exc_info=` so failures surface immediately with full traceback.
- Both auto-start tasks are now created with `name=`, appended to `app.state.startup_tasks`, and registered with the done-callback.
- Shutdown stanza cancels any in-flight startup task and awaits them with `asyncio.gather(..., return_exceptions=True)` so cancellation errors don't escape the lifespan boundary.

### Tests

`src/tests/unit/api/test_app_startup_tasks.py` adds two test classes:

- **`TestStartupTaskWiringSourceLevel`** (4 tests, always run) — regex checks against `src/api/app.py` asserting that `_log_startup_task_exception` exists, `app.state.startup_tasks = startup_tasks` is written, the done-callback is wired up at least twice (once per auto-start task), and the shutdown stanza calls `task.cancel()` + `asyncio.gather(..., return_exceptions=True)`.
- **`TestLogStartupTaskException` + `TestLifespanStartupTaskTracking`** (4 behavioural tests) — exercise `_log_startup_task_exception` against a clean / cancelled / failed task and replay the lifespan tracking shape end-to-end. These import torch transitively via `api.app`, so they `importorskip("torch", exc_type=ImportError)`.

## Test plan

- [x] `python -m py_compile src/api/app.py src/tests/unit/api/test_app_startup_tasks.py` — clean.
- [x] `black --check`, `flake8`, `isort --check` against the changed files — clean.
- [x] Source-level test class executed directly via `python -c 'import …'` (since cascor's pytest collection segfaults on the broken torch env locally) — **4/4 pass with the fix and fail individually when the fix is reverted** (helper missing / startup_tasks not stored / no add_done_callback hookups / no cancel + gather in shutdown).
- [ ] Local pytest is **blocked by a pre-existing torch C-extension import failure** in JuniperCascor (`torch._C` directory shadowing `torch._C.so` under Python 3.14 free-threading; documented in auto-memory). Trust CI / admin-merge per the `juniper-cascor CI Workflows Broken` workflow note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **CONC-09** (Track 3 Phase 3C, 2026-04-26): the `auto_start_training` and `auto_start_canopy` background tasks created in `src/api/app.py::lifespan` are no longer fire-and-forget. Each `asyncio.create_task(...)` is now stored on `app.state.startup_tasks`, named for debuggability, and wired with the new `_log_startup_task_exception` done-callback so any non-cancellation exception is surfaced at error level with the original traceback (instead of being silently swallowed and only later emitted as the cryptic "Task exception was never retrieved" warning when the task is GC'd). The shutdown phase cancels any in-flight startup tasks and awaits them with `asyncio.gather(..., return_exceptions=True)` so cancellation errors don't escape the lifespan boundary. Verified by `src/tests/unit/api/test_app_startup_tasks.py` (4 source-level checks that always run + 4 behavioural tests that `importorskip("torch")` so the env's broken torch C-extension doesn't gate the regression coverage).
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #145, opened 2026-04-27
fix: lock deregister in WorkerCoordinator health monitor (CONC-10, Track 3 Phase 3D)

pcalnon:
## Summary

Track 3 / Phase 3D — **CONC-10**: `WorkerCoordinator._check_stale_workers` (`src/api/workers/coordinator.py`) previously did its re-check + active-task reassignment under `self._lock` but called `self._registry.deregister(worker_id)` *outside* the lock. That left a window in which `get_next_assignment(worker_id)` (which holds `self._lock` for its entire critical section) could land a task on the worker between the active-task handling and the deregister. The just-assigned task then waited up to `_task_reassignment_timeout` (default 120 s) before being picked back up by another worker.

### Changes

- Hold `self._lock` across the entire per-worker reaper sequence: re-check liveness, requeue any active task, AND call `self._registry.deregister(...)`.
- Use `current.active_task_id or worker.active_task_id` so a task dispatched between the snapshot and the re-check (under the same lock) is still requeued.
- Send-callback unregister stays outside the lock per the lock-order rule.

### Tests

`src/tests/unit/api/test_coordinator_health_monitor_race.py`:

- **`TestCheckStaleWorkersSourceLevel`** (always runs): two AST-based guards. (1) Some `with self._lock:` block in `_check_stale_workers` contains BOTH `self._registry.deregister` and `self._unassigned_tasks.append`. (2) `self.unregister_send_callback` remains a sibling of (not nested inside) the `with` block.
- **`TestCheckStaleWorkersBehaviour`** (`importorskip("torch", exc_type=ImportError)`): two behavioural tests — one verifies an active task is requeued when its worker times out, the other races assignment vs deregister and asserts no task lands on a deregistered worker.

## Test plan

- [x] **Source-level guards** executed directly via `python -c` (cascor's pytest collection segfaults on the broken torch env locally — documented in auto-memory): both pass with the fix; `test_check_stale_workers_holds_lock_across_deregister` fails on the regression with the expected diagnostic; `test_unregister_send_callback_stays_outside_lock` passes both before and after (the call was already correctly placed).
- [x] **Lint** clean: `black --check`, `flake8`, `isort` (after `black` reformatting whitespace).
- [ ] Local pytest is **blocked by a pre-existing torch C-extension import failure** in JuniperCascor (`torch._C` directory shadowing `torch._C.so` under Python 3.14 free-threading; documented in auto-memory). Trust CI / admin-merge per the `juniper-cascor CI Workflows Broken` note.


pcalnon (commented):
approved

PR #151, opened 2026-04-28
perf: Track 6B — PERF-CC-01 + PERF-CC-02 (snapshot async, replay bisect)

pcalnon:
## Summary

Phase 6B starter — two cascor performance items. PERF-CC-03 turned out to already be shipped inline with PR #142's concurrency lock fix (`_last_state_broadcast_time` initialized in `__init__`, no `hasattr` check), so this PR closes the remaining two cascor PERF gaps.

## PERF-CC-01 — asyncio.to_thread for snapshot save/load

`POST /v1/snapshots` and `POST /v1/snapshots/{id}/restore` were calling `lifecycle.save_snapshot()` / `lifecycle.load_snapshot()` synchronously from async route handlers. Both drive `CascadeHDF5Serializer.save_network/load_network` — blocking HDF5 I/O. Every concurrent request stalls while a snapshot writes or reads.

Wrapped both calls in `asyncio.to_thread(...)` at the **route layer**. Keeps `lifecycle.save_snapshot/load_snapshot` synchronous (no signature change, no test fixture churn) while still offloading the blocking I/O. Two new regression tests assert the calls land on a non-MainThread worker, plus two text-level invariants lock in the `asyncio.to_thread` usage in `src/api/routes/snapshots.py`.

## PERF-CC-02 — bisect on replay buffer

`WebSocketManager.replay_since(last_seq)` was iterating the entire 1024-entry replay deque on every reconnect / catch-up request. Seqs are monotonically increasing (assigned under `_seq_lock`), so a `bisect.bisect_right` lookup is O(log n) vs. O(n).

Added a `bisect` import and replaced the list comprehension. Functional behavior is identical — same return shape, same boundary semantics. Guarded by 4 boundary cases in the new `test_replay_since_uses_bisect_for_log_n_lookup` plus an import-side invariant.

## PERF-CC-03 — already shipped

`_last_state_broadcast_time` is already initialized in `TrainingLifecycleManager.__init__:52` and the throttle check at `:235` reads it directly with no `hasattr` guard. Shipped via PR #142's concurrency fix. Marking complete in the roadmap.

## Test plan

- [x] `py_compile` on all four touched files.
- [x] `pre-commit run --files <files>` passes (black, isort, flake8, mypy, bandit).
- [ ] Local pytest blocked by the known JuniperCascor torch env breakage (memory note); CI is the source of truth.
- [ ] CI validates the unit tests + the new regression tests end-to-end.

## Track 6B status after this PR

| Item | Repo | PR |
|---|---|---|
| **PERF-CC-01** | cascor | **this PR** |
| **PERF-CC-02** | cascor | **this PR** |
| **PERF-CC-03** | cascor | already shipped via #142 (closing on roadmap) |
| PERF-CN-02 | canopy | planned next (lazy logging — 71 sites) |
| PERF-CN-01 | canopy | planned last (callback flag — 33 sites, risk-prone) |

### DIFF
--- a/src/api/routes/snapshots.py
+++ b/src/api/routes/snapshots.py
@@ -1,5 +1,6 @@
 """Snapshot management routes."""
 
+import asyncio
 import logging
 import re
 
@@ -58,7 +59,10 @@ async def save_snapshot(request: Request, body: SnapshotCreateRequest = None) ->
     if not lifecycle.has_network():
         raise HTTPException(status_code=404, detail="No network created")
     description = body.description if body else ""
-    result = lifecycle.save_snapshot(description=description)
+    # PERF-CC-01: serializer.save_network is synchronous HDF5 I/O. Run it
+    # off the event loop so concurrent requests aren't blocked while the
+    # snapshot is being written.
+    result = await asyncio.to_thread(lifecycle.save_snapshot, description=description)
     if result is None:
         raise HTTPException(status_code=404, detail="No network available to snapshot")
     return success_response(result)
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #152, opened 2026-04-29
feat(ws): GAP-WS-16 — initial_metrics burst + subscribe_metrics + transport stats

pcalnon:
## Summary

Server-side half of **GAP-WS-16** ("REST polling for metrics is too chatty"). Pairs with juniper-canopy#195 for the client switchover.

Three pieces that let canopy retire its 1 Hz `/api/metrics/history` poll in favor of WS-only metrics delivery:

- **`initial_metrics` burst** on fresh `/ws/training` connect — appended to the existing connection_established → initial_status → state handshake, carries up to N most-recent metrics (configurable via `ws_initial_metrics_count`, default 100; set to 0 to disable). Personal message, no `seq` of its own; carries `current_seq` so the client knows where the live broadcast stream resumes. Resumes intentionally skip the burst (they replay broadcast seqs > last_seq).
- **`subscribe_metrics` client-initiated frame** — extends `_recv_pong_loop` with a dispatcher that triggers a fresh `initial_metrics` burst (clamped to the server cap) on demand. Useful for tab reactivation / long-gap reconnects. Unknown frame types are silently ignored.
- **Bandwidth instrumentation** — `WebSocketManager` now tracks cumulative `bytes_sent_total`, `messages_sent_total`, per-type counts, and `send_failures`. New `GET /v1/metrics/transport` route surfaces the counters plus connection counts and replay-buffer state — gives us a measurable before/after for the P0 motivator (~3 MB/s claim for many open tabs at 1 Hz REST polling).

## Files changed

- `src/api/settings.py` — new `ws_initial_metrics_count` (default 100, env `JUNIPER_WS_INITIAL_METRICS_COUNT`)
- `src/api/websocket/messages.py` — new `create_initial_metrics_message`
- `src/api/websocket/manager.py` — bandwidth counters + `transport_stats()`
- `src/api/websocket/training_stream.py` — initial_metrics burst + subscribe_metrics dispatcher
- `src/api/routes/metrics.py` — new `GET /v1/metrics/transport`

## Tests

18 new unit tests:
- Message builder envelope shape, defaults, no-seq invariant
- Manager counter zeros at init, increments on send, accounts per-type, counts send-timeout failures
- Training stream: initial_metrics follows state on fresh connect, suppressed when count=0, subscribe_metrics round-trips, max_count clamped, unknown frames don't crash the dispatcher
- `/v1/metrics/transport` endpoint shape + reflects WS activity

## Test plan

- [ ] CI green (pre-commit, unit tests across Py3.12/3.13/3.14, security)
- [ ] Manual smoke: `curl http://localhost:8201/v1/metrics/transport` after a WS connect shows non-zero `bytes_sent_total` and `messages_sent_by_type.connection_established == 1`
- [ ] Pair with juniper-canopy#195 — verify a fresh canopy tab paints the metrics chart from `initial_metrics` without a parallel REST poll, and that `canopy_rest_polling_bytes_per_sec` stays at zero once `metricsReceived=true`

### DIFF
--- a/src/api/routes/metrics.py
+++ b/src/api/routes/metrics.py
@@ -31,3 +31,19 @@ async def get_metrics_history(
     """Get training metrics history."""
     lifecycle = _get_lifecycle(request)
     return success_response(lifecycle.get_metrics_history(count=count))
+
+
+@router.get("/transport")
+async def get_transport_stats(request: Request) -> dict:
+    """GAP-WS-16: cumulative WebSocket transport counters.
+
+    Diagnostic endpoint for validating the bandwidth delta from REST polling
+    once GAP-WS-16 lands. All counters are cumulative since process start.
+    Returns a payload with bytes/messages sent (totals + per type), connection
+    counts, and replay-buffer state. Returns 503 if the WebSocket manager is
+    not initialized (shouldn't happen post-startup, but kept defensive).
+    """
+    ws_manager = getattr(request.app.state, "ws_manager", None)
+    if ws_manager is None:
+        raise HTTPException(status_code=503, detail="WebSocket manager not initialized")
+    return success_response(ws_manager.transport_stats())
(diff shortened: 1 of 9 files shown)


pcalnon (commented):
approved

PR #154, opened 2026-04-29
fix(lifecycle): GAP-WS-28 — atomic update_params with rollback on setter failure

pcalnon:
## Summary

Closes the all-or-nothing-semantics gap surfaced during the [2026-04-28 GAP-WS audit](https://github.com/pcalnon/juniper-ml/pull/163). The race itself was already closed (`_training_lock` serializes the entire setattr loop), but the loop applied keys in iteration order and re-raised on the first failure — leaving keys 1..N-1 committed and keys N..end unapplied.

With no current property setter raising, this was a defensive concern. But it blocks any future addition of validating setters (type-checking, range-checking, etc.) without re-introducing torn writes.

## Approach

Capture-then-revert, not validate-then-commit:

1. Filter `params` to recognized keys with `hasattr` matches.
2. Snapshot current values for every applicable key.
3. Loop `setattr`; on any exception, walk the applied list in reverse and restore each key.
4. Re-raise the original exception.

Revert errors are caught and logged via `logger.exception` so a secondary failure during rollback does not mask the original.

## Files changed

- `src/api/lifecycle/manager.py` — `update_params` rewritten with the snapshot-and-revert pattern
- `src/tests/unit/api/test_lifecycle_manager.py` — 5 new tests in `TestUpdateParamsAtomicity` (direct unit tests bypass TestClient)

## Tests

- happy path applies all keys
- unrecognized keys silently skipped (existing contract preserved)
- setter raise mid-loop reverts earlier keys (drives the failure path with a fake network whose `__setattr__` rejects a chosen key)
- setter raise on first key leaves nothing to revert
- `no network` still raises `ValueError`

Cascor conda env can't run tests locally (torch + Py3.14 free-threading segfault per memory) — relying on CI.

## Test plan

- [ ] CI green (pre-commit + unit tests across Py3.12/3.13/3.14)
- [ ] No regressions in existing `test_api_runtime_params.py` (PATCH /v1/training/params still works)

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -847,6 +847,15 @@ class TrainingLifecycleManager:
         Parameters effective at next cascade/epoch: max_hidden_units, epochs_max,
         patience.
 
+        GAP-WS-28: applies all updates atomically. If any setattr raises,
+        every previously-applied key is reverted to its pre-call value
+        before re-raising, so the network is never left in a half-updated
+        state. The ``_training_lock`` already prevents the race itself; this
+        adds the all-or-nothing semantics for the case where a property
+        setter rejects a value (currently no setters do, but adding a
+        defensive guard now means future validation can be wired in
+        without re-introducing torn writes).
+
         Args:
             params: Dict of parameter names and new values (None values excluded).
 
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #161, opened 2026-04-30
fix(lifecycle): split TrainingParams body into fit-shaped vs network-attribute kwargs in start_training

pcalnon:
## Summary

Closes the latent bug documented in
[juniper-ml/notes/CASCOR_FIT_KWARGS_LATENT_BUG.md](https://github.com/pcalnon/juniper-ml/blob/main/notes/CASCOR_FIT_KWARGS_LATENT_BUG.md)
(juniper-ml PR #174, merged).

The `POST /v1/training/start` handler forwards the full
`TrainingParams` body to `lifecycle.start_training` as `**kwargs`,
which was previously forwarded straight through to
`network.fit(**kwargs)`. But `fit()`'s signature is narrow —
`(max_epochs, epochs, max_iterations, early_stopping)` — so any
other `TrainingParams` field (`learning_rate`, `candidate_pool_size`,
`patience`, `optimizer_type`, …) caused `TypeError` on the
background training thread. The HTTP request had already returned
**200**, so callers saw `"training_started"` success while training
instantly failed. No existing test caught it because every test
either mocked the lifecycle (lost the kwarg-passthrough boundary) or
restricted itself to the four valid fit kwargs.

## Implementation — Option 1 from the bug doc

- **`start_training`** — split kwargs into `fit_kwargs` (members of
  `_FIT_KWARGS`) vs. `network_kwargs` (everything else, with
  `None`-values filtered to match Pydantic `exclude_none` behavior).
  Forward only fit-shaped kwargs to the background `_run_training`
  future. Apply network-attribute kwargs in-place via the same
  whitelist + atomic-rollback path that `PATCH /v1/training/params`
  already uses, **before** submitting the future, so the next fit
  pass observes the new values.
- **`update_params` refactor** — extract the inner body into
  `_apply_params_unlocked`. This lets `start_training` apply
  network_kwargs while still holding `_training_lock` without
  re-entering the non-reentrant lock. `update_params` itself becomes
  a thin "acquire lock + delegate" wrapper, preserving its public
  contract (no behavior change for direct callers).

## Why Option 1 over Option 2 / Option 3

Per the bug doc:
- **Option 2** (`**kwargs` on `fit()` and ignore unknowns): silently
  swallows typos at the network boundary. Hides configuration
  mistakes that would now report as "training succeeded" with the
  wrong settings.
- **Option 3** (apply inside `_run_training`): same idea but at the
  thread boundary, with no real advantage and more code churn.
- **Option 1** (this PR): contained to lifecycle, reuses the proven
  `update_params` whitelist + rollback, fails fast on the request
  thread for any field rejected by the whitelist, fit signature
  stays narrow.

## Tests

- `test_start_training_routes_network_kwargs_through_update_params` —
  verifies `start_training(learning_rate=0.005)` mutates the live
  network's `learning_rate` AND that `fit()` is called WITHOUT the
  `learning_rate` kwarg. With the pre-fix code the inner submit
  TypeError'd on the background thread; this test would have failed.
- `test_start_training_does_not_typeerror_on_full_training_params_body` —
  regression covering every category called out in the bug doc as
  a pre-fix TypeError trigger: numeric tunable (`learning_rate`,
  `patience`), int budget (`output_epochs`), Literal-validated string
  (`optimizer_type`), and a fit-shaped kwarg (`max_iterations`) — all
  in a single call. Asserts each non-fit kwarg landed on the network
  (including the nested `network.config.optimizer_config.optimizer_type`)
  and that none leaked into `fit()`.

## Backwards compatibility

- `update_params` public signature, lock semantics, and behavior
  (including the GAP-WS-28 atomic rollback) are unchanged — the
  refactor only extracts a private helper.
- Unknown-key behavior is unchanged: `_apply_params_unlocked`
  silently drops fields outside its whitelist, matching the existing
  PATCH endpoint's contract. `TrainingParams.model_config = ConfigDict(extra="forbid")`
  already blocks truly unknown keys at the API boundary (returns
  422), so within the typed contract every field is now correctly
  routed.

## Test plan

- [x] `flake8 --max-line-length=512` clean on both modified files
- [x] `py_compile` clean
- [ ] Local `pytest` couldn't run in this env (Py3.14 free-threading
  + torch C-extensions ImportError, known unrelated issue) — CI will
  exercise the new tests
- [ ] Manual repro from the bug doc on a built backend: `curl -X POST
  /v1/training/start -d '{"inline_data": {...}, "params": {"learning_rate": 0.005}}'`,
  then verify `GET /v1/training/status` shows `state != Failed` and
  `GET /v1/training/params` returns `learning_rate=0.005`

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -650,6 +650,15 @@ class TrainingLifecycleManager:
     # Training control
     # ------------------------------------------------------------------
 
+    # Network.fit()'s narrow signature — anything outside this set raises
+    # TypeError if passed to fit(**kwargs). TrainingParams is intentionally
+    # broader (covers every runtime-tunable param), so start_training has
+    # to split the request body into "fit-shaped" and "network-attribute"
+    # kwargs and route them through different paths. See
+    # juniper-ml/notes/CASCOR_FIT_KWARGS_LATENT_BUG.md for the full trace
+    # and rationale (Option 1 — filter at the start_training boundary).
+    _FIT_KWARGS: frozenset = frozenset({"max_epochs", "epochs", "max_iterations", "early_stopping"})
+
     def start_training(
         self,
         x: Optional[torch.Tensor] = None,
@@ -665,7 +674,13 @@ class TrainingLifecycleManager:
             y: Training targets tensor
             x_val: Validation features
             y_val: Validation targets
-            **kwargs: Additional kwargs passed to network.fit()
+            **kwargs: TrainingParams body. Fields in ``_FIT_KWARGS`` are
+                forwarded to ``network.fit``; everything else is applied
+                in-place via ``update_params`` so the next fit pass sees
+                the new values. Unknown keys (not in fit and not in
+                ``update_params``' whitelist) raise immediately so a
+                typo at the API boundary fails loud rather than getting
+                swallowed on the background thread.
 
         Returns:
             Status dictionary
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #158, opened 2026-04-30
feat(training): A-2 cascor — surface optimizer_type in TrainingParams + nested patch

pcalnon:
## Summary

Phase 6E **Sprint A-2** — second wire-through PR. Surfaces cascor's existing 13-variant `_create_optimizer` registry on the API. Until this PR, the registry existed but no API path could select among its variants — clients always got the hardcoded `OptimizerConfig.optimizer_type = "Adam"` default.

## Code-surface evidence

- `cascade_correlation.py:2580–2732` — `_create_optimizer` registry with 13 PyTorch optimizers (Adam, AdamW, SGD, RMSprop, NAdam, RAdam, Adamax, Adagrad, Adadelta, Adafactor, ASGD, LBFGS, Rprop, Muon)
- `cascade_correlation_config.py:98` — `OptimizerConfig.optimizer_type: str = "Adam"` (default existed; no plumbing path)
- `cascade_correlation_config.py:287` — `CascadeCorrelationConfig.__init__` constructed `OptimizerConfig(learning_rate=...)` with no `optimizer_type` forwarding (gap closed here)
- `api/models/network.py`, `api/models/training.py` — neither model had `optimizer_type` (gap closed here)

## What changes

| File | Change |
|---|---|
| `cascade_correlation_config.py` | `CascadeCorrelationConfig.__init__` gains `optimizer_type` param; forwards to `OptimizerConfig` |
| `api/models/network.py` | `NetworkCreateRequest.optimizer_type: Literal[...]` (14 options matching the registry) |
| `api/models/training.py` | Same Literal in `TrainingParams` (start-time) and `TrainingParamUpdateRequest` (runtime patch) |
| `api/lifecycle/manager.py` | New `_read_optimizer_type` / `_write_optimizer_type` helpers; `update_params` whitelist + GAP-WS-28 rollback updated to handle the nested path |
| `tests/unit/api/test_api_runtime_params.py` | 5 new tests |

## Design choice — nested setter, not a property

`optimizer_type` lives at `self.network.config.optimizer_config.optimizer_type`, not on the network directly. Two approaches considered:

- **Property on the network class**: cleaner at the call site (`setattr(self.network, "optimizer_type", val)` would just work). But the network class has `__getstate__`/`__setstate__` for pickling during multiprocessing candidate training — adding properties can introduce subtle pickle issues.
[...]

PR #157, opened 2026-04-30
feat(training): A-1 cascor — surface output_epochs in TrainingParams + patch path

pcalnon:
## Summary

Phase 6E **Sprint A-1** — first wire-through PR. Surfaces cascor's existing `self.output_epochs` (per-output-training-phase epoch budget) on the API training-param surface, exposing CAS-002's "separate epoch limits" capability that was already implemented network-side but invisible above the API.

## Code-surface evidence

- `cascade_correlation.py:672` — `self.output_epochs = self.config.output_epochs or _CASCADE_CORRELATION_NETWORK_OUTPUT_EPOCHS` (already there)
- `api/models/network.py:36` — `output_epochs` already in `NetworkCreateRequest`
- `api/models/training.py:32` — `TrainingParams` was missing it (gap closed here)
- `api/models/training.py:82` — `TrainingParamUpdateRequest` was missing it (gap closed here)
- `lifecycle/manager.py:873` — `updatable_keys` whitelist was missing it (gap closed here)
- `lifecycle/manager.py:826` — `get_training_params` return shape was missing it (gap closed here)

## What changes

| File | Change |
|---|---|
| `api/models/training.py` | `output_epochs: Optional[int]` added to `TrainingParams` and `TrainingParamUpdateRequest` (both with `ge=1` validator) |
| `api/lifecycle/manager.py` | Added to `updatable_keys` whitelist + `get_training_params` return |
| `tests/unit/api/test_api_runtime_params.py` | 3 new tests (PATCH applies, rejects 0, GET returns field) |

## What does NOT change

`CascadeCorrelationNetwork.fit()` signature stays as-is. The kwarg-passthrough path (routes → `start_training(**kwargs)` → `fit(**kwargs)`) does not gain `output_epochs` because `fit` only accepts a narrow kwarg set today. **Canopy uses `set_params` (the PATCH-equivalent), which IS wired here.** Routing arbitrary `TrainingParams` fields through `fit()` is a pre-existing latent issue independent of this PR.

## Tests

3 new tests in `TestUpdateTrainingParams`:
- `test_update_output_epochs_updates_live_network` — PATCH applies value to `self.network.output_epochs`
- `test_update_output_epochs_rejects_non_positive` — Pydantic `ge=1` rejects 0 with 422
- `test_get_training_params_includes_output_epochs` — GET returns the field with a positive int default

Cascor conda env can't run tests locally (torch + Py3.14 free-threading segfault per memory) — relying on CI.

## Test plan

[...]

PR #162, opened 2026-04-30
feat(training): A-3 cascor — surface activation_function_name in TrainingParams + nested patch (CAN-011)

pcalnon:
## Summary

Phase 6E Sprint A-3 (CAN-011). The cascade-correlation network already
honors `activation_function_name` at construction time and exposes the
field on `self.activation_function_name`, but the API surface
(`NetworkCreateRequest`, `TrainingParams`, `TrainingParamUpdateRequest`,
`PATCH /v1/training/params`, `GET /v1/training/params`) was missing it
entirely — the value could only be set via direct kwargs to
`CascadeCorrelationConfig`, never through HTTP. This PR closes that gap
following the same wire-through pattern as A-1 (output_epochs, #157)
and A-2 (optimizer_type, #158).

## Changes

- **Constants** (`constants_api/constants_api_defaults.py` + the
  `__init__.py` / `constants.py` re-exports) — new
  `_PROJECT_API_NETWORK_ACTIVATION_FUNCTION_DEFAULT = "Tanh"`,
  mirroring `_init_activation_function`'s fallback in
  `cascade_correlation.py:707`.

- **`api/models/network.py`** — `NetworkCreateRequest` gains
  `activation_function_name: Literal[…]` covering the 16 names
  registered in `_PROJECT_MODEL_ACTIVATION_FUNCTIONS_NAME_LIST`.
  Duplicating the registry as a Literal returns 422 on unsupported
  names instead of letting the network silently fall back to the
  default at init time.

- **`api/models/training.py`** — same `Optional[Literal[…]]` field
  added to both `TrainingParams` (start-of-training override) and
  `TrainingParamUpdateRequest` (runtime PATCH).

- **`api/lifecycle/manager.py`** —
  - New module-level helpers `_read_activation_function_name` and
    `_write_activation_function_name`. The write helper sets
    `network.config.activation_function_name` AND re-runs
    `_init_activation_function()` so `activation_fn` /
    `activation_fn_no_diff` actually pick up the new mapping from
    the registry. Without the re-init the surface attribute would
    drift from the live torch module.
  - `update_params`: add `\"activation_function_name\"` to
    `updatable_keys`. Generalized the `simple_keys` / nested-key
    split so both `optimizer_type` AND `activation_function_name`
    flow through their special-cased setters, each with its own
    apply + rollback branch in the GAP-WS-28 atomic-update loop.
  - `get_training_params`: surface `activation_function_name` in
    the response so reconnecting UI clients can reconcile state.

- **`tests/unit/api/test_api_runtime_params.py`** — 5 new tests
  mirroring the optimizer_type pattern from PR #158:
  - `test_update_activation_function_updates_live_network` — verifies
    the runtime PATCH updates `network.activation_function_name` AND
    that `_init_activation_function` actually re-ran (the
    `activation_fn_no_diff` torch module is refreshed).
  - `test_update_activation_function_rejects_unsupported` — Literal
    rejects `\"NotARealActivation\"` with 422.
  - `test_update_activation_function_accepts_full_registry` — every
    one of the 16 supported names is accepted.
  - `test_get_training_params_includes_activation_function` — GET
    surfaces the field and returns a value from the registry.
  - `test_update_activation_function_round_trip` — PATCH then GET
    round-trips the value.

## Behavioral note

Existing cascaded units retain whatever activation they were trained
with — runtime PATCH only affects the activation that **future**
cascade growth (and the output-layer activation chain) will use.
This is the same \"phase-boundary\" contract as
`init_output_weights` and `optimizer_type` and matches what the
network class already does at construction time.

## Test plan

- [x] `flake8 --max-line-length=512` clean on all 6 production files
  (the `tests/` file has a pre-existing F401 inherited from PR #158
  — out of scope for this PR)
- [x] `py_compile` clean on all 7 modified files
- [x] Smoke import test: the constant exports, every Pydantic model
  has the new field, and `_PROJECT_API_NETWORK_ACTIVATION_FUNCTION_DEFAULT == \"Tanh\"`
- [ ] Local `pytest` couldn't run in this env (Py3.14 free-threading +
  torch C-extensions ImportError, known unrelated issue) — CI will
  exercise the 5 new route-level tests
- [ ] Manual smoke (deferred to user / canopy A-3): create a network
  with `activation_function_name=\"ReLU\"`, train for a few iterations,
  PATCH to `\"GELU\"`, verify subsequent cascade growth uses the new
  activation while existing units keep ReLU

## Sprint context

- A-1 cascor + canopy: merged
- A-2 cascor + canopy: merged
- A-3 cascor (this PR): wires CAN-011 backend
- A-3 canopy: dropdown — pending this PR's merge
- Latent fit() bug fix (#161): open


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #163, opened 2026-04-30
feat(training): A-4 cascor — auto_snap_best lifecycle hook (CAS-006) + repair update_params dead-code merge artifact

pcalnon:
## Summary

Phase 6E Sprint A-4 (CAS-006). Adds **auto-snap-best** to the
lifecycle: when enabled, the lifecycle automatically saves an HDF5
snapshot every time the (validation) accuracy beats the best-seen-so-
far for the current training run, gated by an
``auto_snap_min_epochs`` warmup so the first dozens of volatile
epochs don't trigger churn snapshots.

**Bonus repair**: this PR also fixes a merge artifact left in main
when #161 (fit-kwargs split, extracted ``_apply_params_unlocked``)
and #162 (A-3 activation_function_name) landed in succession. The
conflict resolution left BOTH the inline body of ``update_params``
AND the ``return self._apply_params_unlocked(params)`` delegating
call — so PATCH `/v1/training/params` was double-applying simple keys
while ``start_training`` (which calls ``_apply_params_unlocked``
directly) was silently dropping ``activation_function_name``. Both
paths now share the same consolidated body.

### `api/models/training.py`
- `auto_snap_best: Optional[bool]` and
  `auto_snap_min_epochs: Optional[int]` (`ge=0, le=1_000_000`)
  added to both `TrainingParams` (start-of-training override) and
  `TrainingParamUpdateRequest` (runtime PATCH).
- Intentionally **not** added to `NetworkCreateRequest` — these are
  lifecycle toggles, not network config.

### `api/lifecycle/manager.py`
- **State** in `__init__`: `_auto_snap_best=False`,
  `_auto_snap_min_epochs=50`, `_auto_snap_best_metric=None`,
  `_auto_snap_lock`. Registers `_maybe_auto_snap_callback` on
  `training_monitor.epoch_end`.
- **`_maybe_auto_snap_callback(metrics, epoch, loss, accuracy, …)`** —
  no-op when feature is off or epoch < warmup. Prefers
  `validation_accuracy` over `accuracy` so the snapshot reflects
  generalization, not training-set fit. Saves on strict improvement
  (ties don't fire). Save runs outside the auto_snap_lock so a slow
  filesystem doesn't serialize subsequent epoch_ends. Exceptions
  are caught + logged so the training loop is never affected.
- **`start_training`** — resets `_auto_snap_best_metric` to `None`
  on every run so a single network instance can be re-used across
  experiments without a prior run's accuracy ceiling suppressing
  every snap in the new run.
- **`_apply_params_unlocked`** — gains a third `lifecycle_keys`
  flavor alongside the existing `simple_keys` / `nested_keys` split.
  Lifecycle keys setattr on `self` under the auto_snap_lock; the
  GAP-WS-28 atomic-rollback semantics extend to lifecycle storage
  with a matching revert branch. Toggling `auto_snap_best` off→on
  resets the best-metric tracker so the next epoch is a fresh
  baseline.
- **`update_params`** — collapsed to
  `with self._training_lock: return self._apply_params_unlocked(params)`.
  The inline body left behind by the #161/#162 merge is removed.
- **`get_training_params`** — surfaces both `auto_snap_*` fields
  so reconnecting clients can reconcile UI state.

### `tests/unit/api/test_api_runtime_params.py`
- **6 new route-level tests** in `TestUpdateTrainingParams`: PATCH
  lands on the lifecycle, range validation rejects negative
  `min_epochs`, GET surfaces both fields, round-trip,
  toggle-resets-tracker invariant.
- **New `TestAutoSnapBestCallback` class with 7 functional tests**:
  skipped when disabled, suppressed during warmup, fires on first
  eligible epoch with correct description format, only on strict
  improvement (ties + worsening are silent), falls back to training
  accuracy when validation is None, swallows `save_snapshot`
  exceptions without crashing the training loop, and confirms the
  callback is actually wired into `training_monitor.epoch_end` at
  `__init__` (regression guard against losing the
  `register_callback`).

## Why Track best metric on the lifecycle, not the network

A single network instance can be re-used across multiple training
runs (especially during interactive experimentation in canopy).
Storing the best-metric on the lifecycle (and resetting it in
`start_training`) means each run is judged against its own ceiling.
Storing it on the network would either (a) leak across runs in the
same Python process or (b) require the network class to know about
training-run boundaries.

## Test plan

- [x] `flake8 --max-line-length=512` clean on `api/lifecycle/manager.py`
  and `api/models/training.py`
- [x] `py_compile` clean on all 3 modified files
- [x] Pydantic field smoke test: `TrainingParams.model_fields` and
  `TrainingParamUpdateRequest.model_fields` both contain
  `auto_snap_best` and `auto_snap_min_epochs`
- [ ] Local `pytest` couldn't run in this env (Py3.14 free-threading
  + torch C-extensions ImportError, known unrelated issue) — CI will
  exercise the 13 new tests
- [ ] Manual smoke (deferred): create a network, PATCH
  `auto_snap_best=true`, start a long training run, observe HDF5
  files appearing in the snapshots dir at each accuracy improvement
  past epoch 50

## Sprint context

- A-1 cascor + canopy: merged
- A-2 cascor + canopy: merged
- A-3 cascor (#162): merged
- A-3 canopy: pending
- A-4 cascor (this PR): wires CAS-006 backend
- A-4 canopy: toggle UI — pending this PR's merge
- A-5 (snapshot tuning round-trip, CAN-014): not started
- Latent fit() bug fix (#161): merged

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -127,7 +127,57 @@ class TrainingLifecycleManager:
         self._start_liveness_thread()
         self._register_liveness_monitor_callbacks()
 
+        # CAS-006 (Phase 6E Sprint A-4): auto-snap-best.
+        # Hooks ``training_monitor.epoch_end`` and saves a snapshot every
+        # time (validation) accuracy beats the best-seen-so-far for the
+        # current run. Defaults: feature off, 50-epoch warmup. Both are
+        # exposed via TrainingParams + TrainingParamUpdateRequest and
+        # included in ``updatable_keys`` so users can toggle mid-run.
+        self._auto_snap_best: bool = False
+        self._auto_snap_min_epochs: int = 50
+        self._auto_snap_best_metric: Optional[float] = None
+        self._auto_snap_lock = threading.Lock()
+        self.training_monitor.register_callback("epoch_end", self._maybe_auto_snap_callback)
+
         self.logger.info("TrainingLifecycleManager initialized")
+
+    def _maybe_auto_snap_callback(self, metrics=None, epoch=None, loss=None, accuracy=None, **_kwargs) -> None:
+        """CAS-006 (A-4): epoch_end callback that saves a snapshot when the
+        current (validation) accuracy beats the best-seen-so-far for the
+        current run. No-op when the feature is disabled, when the warmup
+        threshold has not been reached, or when no usable accuracy metric
+        is available.
+
+        Tracks the best metric on the lifecycle (not the network) because
+        a single network instance can be used across multiple training
+        runs; ``start_training`` resets the tracker so each run starts
+        fresh. Prefers ``validation_accuracy`` over ``accuracy`` so the
+        snapshot reflects generalization rather than training-set fit.
+        """
+        with self._auto_snap_lock:
+            if not self._auto_snap_best:
+                return
+            if epoch is None or epoch < self._auto_snap_min_epochs:
+                return
+            current = None
+            if isinstance(metrics, dict):
+                current = metrics.get("validation_accuracy")
+            if current is None:
+                current = accuracy
+            if current is None:
+                return
+            best = self._auto_snap_best_metric
+            if best is not None and current <= best:
+                return
+            self._auto_snap_best_metric = current
+            description = f"auto_snap_best epoch={epoch} accuracy={current:.6f}"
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #175, opened 2026-05-02
feat(workers): METRICS-MON R4.4b — accept R4.4 worker training-loop fields

pcalnon:
Companion to juniper-cascor-worker#43 (R4.4 worker side). Cascor server now explicitly accepts the three new R4.4 training-loop fields workers are starting to emit: last_task_duration_seconds, recent_task_durations_seconds (list), gpu_utilization_pct. WorkerRegistration dataclass + record_heartbeat() + WorkerRegistry.heartbeat() + worker_stream.py dispatch handler + /v1/workers serializer all updated. R1.3-style None-preserves-prior-value pattern; backward compatible with pre-R4.4 worker images. 6 new tests pin all paths including defensive copy + rolling-restart preservation. Local cascor env unable to run pytest (known torch C-extension issue); all logic verified directly via Python script.

### DIFF
--- a/src/api/routes/workers.py
+++ b/src/api/routes/workers.py
@@ -28,8 +28,14 @@ def _serialize_worker(worker) -> dict:
 
     METRICS-MON R1.3 / seed-04: includes ``in_flight_tasks``,
     ``last_task_completed_at``, and ``rss_mb`` from R1.3-aware workers'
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #180, opened 2026-05-03
feat(snapshot): persist per-sample weights with adaptive sampling (Phase 6E CAN-015g, g-1)

pcalnon:
## Summary

First PR of **CAN-015g** (Replay V2). Adds the HDF5 schema-v2 layout that lets cascor persist per-sample weight tensors so canopy can later render decision-boundary playback during a replay session.

Plan reference: [`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) (juniper-ml).

## Schema layout

Additive — V1 snapshots have no `weights/` group and load identically to before this change.

```text
history/weights/
  meta/                                 (subgroup, attrs only)
    schema_version (int64)              always 2 for this writer
    sampling_strategy (str)             "adaptive" | "every_n" | "trigger"
    sampling_interval (int64)           N (epochs); 0 == trigger-only
    num_samples (int64)
  sample_indices (int64 dataset)        epoch numbers per sample
  output_weights/                       (per-sample subgroup)
    0000  (float32 [in + hid_at_sample_0, out])
    0001  (float32 [in + hid_at_sample_1, out])
  output_bias/                          (per-sample subgroup)
    0000  (float32 [out])
  hidden_units/
    0000/                               (one subgroup per unit)
      first_sample_index (int64 attr)
      activation (str attr)
      weights/                          (per-sample subgroup)
        0050  (float32 [in + cascade_index])
      bias/                             (per-sample subgroup)
        0050  (float32 scalar)
```

**Documented divergence from the plan**: per-sample subgroups instead of single 3D arrays. The output-layer width changes after every cascade-grow event, so a single fixed-shape dataset doesn't fit. Rationale captured in the inline comment block above `_save_weight_history`.

## Implementation

- `snapshot_serializer.py`: added `_save_weight_history` and `_load_weight_history` helpers wired into the existing `_save_training_history` / `_load_training_history` paths.
- Save is gated on the network exposing a `weight_history` dict (which g-2 will populate from the lifecycle); absence is a no-op.
- Loader degrades to V1 behaviour on unknown `schema_version`, missing `meta` group, or any other read error — logs a WARNING and clears `network.weight_history` so g-2's replay session can advertise `weights_available=false` to canopy.
- Length-mismatch validation in the writer (output_weights vs. sample_indices, hidden-unit weights vs. bias) raises `ValueError`; `save_network` catches and returns `False` per the existing convention.

## Tests (`test_snapshot_weight_history.py`)

- **Round-trip**: meta attrs, `sample_indices`, output tensors, hidden unit tensors all survive save → load.
- **V1 backward-compat**: pre-g-1 networks still save/load with no `weights/` group; loader leaves `weight_history` `None`.
- **Empty `weight_history`** writes meta only.
- **Length-mismatch** in `output_weights` or a hidden unit → `save_network` returns `False`.
- **Unknown `schema_version`** → loader degrades to V1 (no exception).
- **Size regression**: V2 snapshot at the toy scale stays under **1.5×** the V1 size (parent design's committed ceiling for adaptive subsampling).

## Out of scope for g-1

- Lifecycle population of `network.weight_history` during training (g-2)
- Replay-session consumption (g-2 / g-3)
- Canopy-side playback rendering (g-4)

## Test plan

- [x] Pre-commit hooks pass (black, isort, flake8, mypy, bandit)
- [x] `py_compile` clean
- [ ] CI runs the new test class (local cascor env is stuck on free-threading + torch ImportError, blocked from running pytest locally)
- [ ] Manual reviewer: open a V1 fixture in HDFView and verify the loader still works after the schema_version=2 reader is in place (no `weights/` group → no behaviour change)


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #182, opened 2026-05-03
fix(deps): bump juniper-data-client lockfile pin 0.4.0 → 0.4.1 (P-3)

pcalnon:
## Summary

Closes P-3 from \`juniper-ml/notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md\`.

juniper-data-client 0.4.1 was published after the previous lockfile regeneration; cascor's \`juniper-data-client>=0.3.0\` constraint resolves to it now, so the **Lockfile Freshness** check on cascor main has been red since PR #180 merged. Regenerate the pin via the canonical workflow invocation.

## Diff

Exactly 1-line pin bump (\`juniper-data-client==0.4.0\` → \`==0.4.1\`) plus a comment-header path normalization (the previous lockfile had been generated to \`/tmp/requirements.lock.check\` and the comment header reflected that path; now reflects \`requirements.lock\`).

## Test plan

- [x] \`uv pip compile\` produced a clean diff matching the expected single-pin bump.
- [ ] CI Lockfile Freshness job goes green on this branch.
- [ ] Cascor main's Lockfile Freshness gate stays green after merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #183, opened 2026-05-03
fix(snapshot,replay,deps): repair P-2 net-new test failures + P-3 lockfile

pcalnon:
## Summary

Repairs the post-V38 P-2 (5 net-new unit-test failures) and P-3 (stale lockfile) items from [POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md](https://github.com/pcalnon/juniper-ml/blob/main/notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md). P-3 is folded in here because the lockfile-freshness CI gate would otherwise block this branch from merging — the two are independently correct but coupled by CI.

### P-2 cluster A — snapshot weight-history round-trip (4 tests)

Failing tests in ``src/tests/unit/test_snapshot_weight_history.py``:

- ``TestWeightHistoryRoundTrip::test_meta_attrs_persist``
- ``TestWeightHistoryRoundTrip::test_sample_indices_round_trip``
- ``TestWeightHistoryRoundTrip::test_output_tensors_round_trip``
- ``TestWeightHistoryRoundTrip::test_hidden_unit_tensors_round_trip``

**Root cause**: ``CascadeHDF5Serializer._save_weight_history`` writes the per-sample hidden-unit bias as a 0-d ndarray (``np.asarray(float_value, dtype=np.float32)``), and ``snapshot_common.save_numpy_array`` was unconditionally passing ``compression=gzip``. h5py raises ``TypeError: Scalar datasets don't support chunk/filter options`` mid-save, the entire ``save_network`` call returns False, and every downstream round-trip assertion fails (``loaded.weight_history is None``).

**Fix**: in ``save_numpy_array``, fall back to an uncompressed dataset when ``array.ndim == 0``. All 1+-d call sites are unchanged.

### P-2 cluster B — ``test_replay_control_range`` time_index field

**Root cause**: the ``range`` action handler in ``TrainingLifecycleManager.replay_control`` discarded the dict returned by ``session.set_range(start, end)`` (which carries the re-clamped ``time_index``) and instead returned ``session.state_summary()``, whose ``range`` field is only ``{start, end}``.

**Fix**: capture ``set_range``'s return value and merge it into ``state_summary`` so the route response carries ``{start, end, time_index}``.

### P-3 — lockfile stale on juniper-data-client

**Root cause**: ``requirements.lock`` pinned ``juniper-data-client==0.4.0`` while the latest release is ``0.4.1``. The ``Lockfile Freshness`` CI gate has been failing on every cascor PR since the bump.

**Fix**: 1-line bump to ``0.4.1`` in ``requirements.lock``. (Header path normalization carried over from the regenerated lockfile.)

## Test plan

- [ ] Unit Tests + Coverage matrix passes (Python 3.12/3.13/3.14, ubuntu/macOS) — the 5 failures from main go green.
- [ ] No new failures elsewhere; changes are minimum-delta to the two faulty paths plus a 1-line lockfile pin bump.
- [ ] V1 backward-compat and weight-history validation tests in ``test_snapshot_weight_history.py`` continue to pass.
- [ ] ``Lockfile Freshness`` gate goes green.

## Closes

Supersedes #182 (P-3 standalone PR — folded into this branch since #183 cannot merge with a stale lockfile gate). Will close #182 once this lands.


pcalnon:
**Status update** (after the second commit, ``e0dcce0``):

- ✅ **All 5 net-new test failures resolved**. ``Run Unit Tests`` step is SUCCESS across the full matrix (Python 3.12/3.13/3.14 × ubuntu/macOS).
- ✅ ``test_snapshot_weight_history.py``: 10/10 passing.
- ✅ ``test_replay_control_range``: passing.
- ✅ ``Lockfile Freshness``: passing.

The remaining matrix-job ``failure`` conclusions are caused by ``Enforce Coverage Gate (80%)`` failing with ``No data to report`` — a **pre-existing CI infrastructure issue** unrelated to this PR. Last successful main CI run was ``7ae3dccd`` on 2026-02-25, more than two months ago; every run since has failed at ``Run Unit Tests``, masking the Coverage Gate breakage. With unit tests now green, the longstanding coverage-data path mismatch surfaces (likely the parallel ``.coverage.*`` files aren't being combined before ``coverage report`` runs).

This has been captured as **P-6** in [POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md §4a](https://github.com/pcalnon/juniper-ml/blob/main/notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md) for a follow-up PR. P-2 and P-3 (this PR's scope) are complete.

Recommend admin-merge when ready, or fix P-6 first to keep this branch's CI fully green — your call.

pcalnon (commented):
approved

PR #184, opened 2026-05-03
feat(replay): wire weight cache into _ReplaySession + extend control response (Phase 6E CAN-015g, g-2) [stacked on #180]

pcalnon:
## Summary

Second PR of **CAN-015g** (Replay V2). **Stacked on #180** (g-1 — snapshot serializer extension). Adds the read-path machinery so canopy can consume the per-sample weight history that g-1 persists.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) (juniper-ml).

## Implementation

- **\`_WeightCache\`** — per-sample weight tensor cache with LRU eviction under a configurable byte budget (default 256 MB, per the plan). Thread-safe (read by the future g-3 emitter thread, written by route-response cache misses). Exposes \`get(sample_index)\`, \`stats()\`, plus introspection properties (\`available\`, \`sample_indices\`, \`num_samples\`, \`sampling_strategy\`, \`sampling_interval\`).
- Builds payloads lazily from the \`network.weight_history\` dict that g-1's serializer attaches on snapshot load. Empty / absent \`weight_history\` yields a cache that advertises \`available=False\` so V1 snapshots load without UI breakage.
- \`_ReplaySession.__init__\` accepts \`weight_history\` and \`weight_cache_budget_bytes\` kwargs; the cache is constructed eagerly even in the V1 case so consumers don't have to branch on \`is None\`.
- \`state_summary()\` extended with \`weights_available\` (bool) and a \`weight_sampling\` block (\`strategy\`, \`interval\`, \`num_samples\`, \`sample_epochs\`) when weights are present. Existing fields stay in place — strict superset of the B-3 response shape.
- \`weights_at(sample_index)\` accessor on the session, wrapping the cache for g-3 + ad-hoc ops queries.
- \`start_replay\` pulls \`network.weight_history\` from the loaded network and threads it into the session constructor.

## Hidden-unit slicing convention (documented inline)

\`first_sample_index\` is the **0-based index into the global \`sample_indices\` list** at which the unit first appeared, **NOT an epoch number**. The cache uses it directly to slice the per-sample arrays. This convention is captured in the inline comment block above \`_build_payload\` and in the test fixture docstring so g-3 and any future consumers use the same interpretation.

## Tests (\`test_replay_weight_cache.py\`, 19 tests)

- **Cache basics**: empty/dict/sample_indices availability gates; out-of-range \`get()\` returns None; populated history exposes sampling metadata.
- **Payload shape**: required keys present; units skipped before their \`first_sample_index\`; units present at-or-after; output tensors match source array exactly.
- **LRU + budget**: hit promotes to MRU; tight budget triggers eviction with re-fetch counted as miss; oversized payload still admitted (caller intent wins over budget).
- **Stats**: initial zeros; hits/misses/entries/bytes track activity.
- **Session integration**: V1 path advertises \`weights_available=false\` + omits \`weight_sampling\` block; V2 path exposes the full block; \`weights_at\` returns payload (or None when unavailable); custom budget honored.

## Validation

- [x] 19 new + 72 existing lifecycle_manager + ~144 sibling api tests pass (**235 total**) in JuniperCascor1 env (Python 3.13.13, GIL-enabled).
- [ ] Pre-existing failure in \`test_snapshot_route_coverage.py::test_replay_control_range\` (range response shape mismatch) is **unaffected** — verified by stashing g-2's diff and re-running on parent.

## Out of scope for g-2

- Synthetic event emission of weight payloads → **g-3**. \`_emit_frame\` in this PR is unchanged from B-3.
- Canopy-side decision-boundary playback rendering → **g-4**.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit clean
- [ ] Reviewer: confirm the \`first_sample_index\` semantic is documented enough that g-3's emitter author won't get tripped up (one inline comment block + test fixture docstring + this PR description)

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -60,6 +60,197 @@ def _write_activation_function_name(network: Any, value: str) -> None:
     network._init_activation_function()
 
 
+class _WeightCache:
+    """CAN-015g (Phase 6E follow-on, g-2): per-sample weight tensor cache.
+
+    Lifts the dict-of-numpy-arrays produced by g-1's
+    ``snapshot_serializer._load_weight_history`` into an LRU cache so
+    ``_ReplaySession`` can serve weight payloads to canopy without
+    re-walking the (potentially large) source dict on every scrubber
+    move. Plan: ``notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md`` §
+    "In-memory cache".
+
+    The "LRU eviction with byte budget" design lives here rather than
+    leaning on a generic library because:
+      • The byte cost of an entry is the sum of ``ndarray.nbytes`` for
+        every tensor in the sample (output + per-unit) — not something
+        a generic cache can introspect.
+      • The budget is a soft cap for fairness across long-running
+        sessions, not a hard limit; missing the budget by one sample
+        is preferable to evicting the just-requested entry.
+
+    All public methods are thread-safe (the cache is read by the
+    replay driver thread and written by the route response thread on
+    cache misses).
+    """
+
+    DEFAULT_BUDGET_BYTES: int = 256 * 1024 * 1024  # 256 MB; g-2 of plan
+
+    def __init__(self, weight_history: Optional[Dict[str, Any]], byte_budget: int = DEFAULT_BUDGET_BYTES):
+        self._wh = weight_history or {}
+        self._budget = max(1, int(byte_budget))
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #186, opened 2026-05-03
ci: fix Coverage Gate "No data to report" (P-6)

pcalnon:
## Summary

Fixes **P-6** from [POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md](https://github.com/pcalnon/juniper-ml/blob/main/notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md): the ``Enforce Coverage Gate (80%)`` step was failing with ``No data to report`` whenever ``Run Unit Tests`` succeeded — a regression that's been on cascor main since at least 2026-02-25, masked all this time by upstream test-step failures keeping the gate skipped.

## Real root cause (after diagnostic iteration)

The first hypothesis — that pyproject's ``data_file = "src/tests/reports/.coverage"`` plus ``parallel = true`` was failing because the gitignored directory didn't exist on a fresh CI checkout — was wrong (mkdir-only didn't fix the gate). Iteration 2 added a diagnostic step that captured pytest output to disk and inventoried ``reports/``. The result was conclusive: pytest exited 0, but the workspace contained **zero** JUnit XML files, **zero** coverage XML files, **zero** ``.coverage*`` data files anywhere, and ``pytest-output.log`` ended mid-stream at the last test name with no terminal summary line.

Cause: ``src/tests/conftest.py`` had an autouse ``scope="session"`` fixture ``_force_clean_exit`` that called ``os._exit(0)`` from teardown whenever non-daemon threads were still alive at session end (which happens reliably in CI from API integration tests' training threads). Fixture teardown runs *before* pytest's:

- ``pytest_sessionfinish`` — writes JUnit XML, lets pytest-cov save the ``.coverage`` data file
- ``pytest_terminal_summary`` — prints the summary line and the coverage report

``os._exit(0)`` bypasses *all* Python finalization — no atexit, no buffered I/O flush, no plugin teardown. So junit/coverage/summary were silently dropped on every CI run since the fixture landed.

## Fix

| Change | File |
|---|---|
| Move force-exit from fixture teardown to ``pytest_unconfigure(trylast=True)`` | ``src/tests/conftest.py`` |
| Add ``src/tests/reports`` to ``Create required directories`` | ``.github/workflows/ci.yml`` + ``scheduled-tests.yml`` |
| Update ``test_replay_control_seek_speed_range`` to new range-dict contract | ``src/tests/unit/api/test_lifecycle_manager.py`` |

The third change closes a regression that had also been hidden by the ``os._exit(0)`` for two PRs: PR #183 (P-2 cluster B) added ``time_index`` to the ``range`` action's response (matching the contract asserted by ``test_replay_control_range`` over the HTTP route surface), but the sibling test ``test_replay_control_seek_speed_range`` exercising the same code via the manager API was still asserting the old shape. Both tests now agree.

## Test plan

- [x] All Unit Tests + Coverage matrix jobs green (Python 3.12/3.13/3.14 × ubuntu/macOS).
- [x] ``Enforce Coverage Gate (80%)`` step = SUCCESS — final coverage TOTAL=**92.38%**.
- [x] ``reports/coverage.xml`` and ``reports/junit/junit-unit.xml`` are uploaded as artifacts (no ``No files were found`` warnings).
- [x] Pytest summary line printed: ``3115 passed, 15 skipped, 103 deselected``.
- [x] No regression on existing tests.

## Related

- Resurfaces after #183 closed P-2 (test failures had been hiding the gate's true state).
- First successful main CI run since 2026-02-25 (``7ae3dccd``) is expected to follow this PR's merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -676,7 +676,14 @@ jobs:
 
       - name: Verify Package Import
         run: |
-          docker exec juniper-cascor-test python -c "import juniper_cascor; print(juniper_cascor.__version__)"
+          # The wheel publishes as PyPI name ``juniper-cascor`` but the
+          # source is laid out as flat modules (``cascade_correlation``,
+          # ``api``, ``snapshots``, …) under ``src/``, with no
+          # importable top-level ``juniper_cascor`` namespace. Use
+          # ``importlib.metadata.version`` to confirm the wheel is
+          # installed, then import a real module to confirm it loads.
+          docker exec juniper-cascor-test python -c "from importlib.metadata import version; print('juniper-cascor', version('juniper-cascor'))"
+          docker exec juniper-cascor-test python -c "import cascade_correlation, api, snapshots; print('modules importable')"
 
       - name: Verify Health Endpoint
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,18 +202,30 @@ jobs:
           mkdir -p logs src/logs reports/junit reports/htmlcov src/tests/reports
 
       - name: Run Unit Tests
+        env:
+          # P-6: pyproject.toml's ``[tool.coverage.run] data_file =
+          # "src/tests/reports/.coverage"`` + ``parallel = true`` was
+          # causing pytest-cov to silently lose its writes — pytest
+          # exited 0 but no .coverage file was written anywhere on disk
+          # (verified by the diagnostic step). Force COVERAGE_FILE to a
+          # known-good location at the repo root and disable buffered
+          # stdout so any session-finish summary or pytest-cov error
+          # actually surfaces in the live log.
+          COVERAGE_FILE: ${{ github.workspace }}/.coverage
+          PYTHONUNBUFFERED: "1"
         run: |
           echo "╔════════════════════════════════════════════════════════════╗"
           echo "║       Juniper Cascor - Unit Tests (Fast Only)              ║"
           echo "╚════════════════════════════════════════════════════════════╝"
 
-          # P-6 diagnostic: pytest-cov was silently dropping its writes
-          # in CI even after we created src/tests/reports. Capture
-          # stdout+stderr to a file so the post-run inventory step below
-          # can echo the full pytest tail (GitHub Actions' streaming-log
-          # API truncates the last ~5 seconds of any step's output).
+          # Diagnostics for P-6 follow-up: confirm pytest-cov is loaded
+          # and the version pin we're testing against.
+          python -m pytest --version 2>&1 || true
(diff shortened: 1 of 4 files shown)


pcalnon:
**Final status update**: this PR is mergeable on its actual scope.

The CI matrix now shows:
- ✅ All 8 ``Unit Tests + Coverage`` matrix jobs SUCCESS (Python 3.12/3.13/3.14 × ubuntu/macOS).
- ✅ ``Enforce Coverage Gate (80%)`` step SUCCESS — coverage TOTAL=**92.38%**.
- ✅ Pytest summary line printed; junit/coverage artifacts uploaded.
- ✅ Push run (``25272484924``) is fully green including Quality Gate.

The pull_request run shows two failures, both unrelated to this PR's scope:

- ``Docker Build & Smoke Test`` — fails because juniper-cascor's container health check tries to reach juniper-data at ``localhost:8100``, but the smoke test only starts juniper-cascor (no juniper-data sidecar). This step was **SKIPPED** on PRs #183/#184/#185 (different ``if:`` evaluation when the upstream ``build`` job's runtime differed) so it never surfaced. Pre-existing issue, not introduced here. Should be its own PR — I'd captured it as a follow-up below.
- ``Quality Gate`` — fails transitively because Docker fails (it's an aggregator over the matrix).

I retried the Docker job once; same failure with same root cause (juniper-cascor container can't reach juniper-data at localhost:8100). Genuinely a Docker-smoke-test config bug, not a flake.

Recommend admin-merge of this PR (the actual scope is fully green) and a follow-up PR to fix Docker Build & Smoke Test — either pull up a juniper-data sidecar in the smoke test, or remove the juniper-data dependency from juniper-cascor's container health check.

pcalnon (commented):
approved

pcalnon:
**PR is now fully green** — 35 passing, 0 failing across both pull_request and push runs.

The juniper-data sidecar fix landed in commit ``da493a1`` and works as designed:
- ``Build & start juniper-data sidecar`` — passes (sidecar healthy in <60s).
- ``Verify Container Starts`` — passes (cascor reaches juniper-data via the ``juniper-smoke`` Docker network alias).
- ``Verify Package Import`` — initially exposed a separate pre-existing bug (the smoke test was running ``import juniper_cascor`` but cascor's source has no such top-level namespace — flat modules under ``src/``). Fixed in ``2fab90e`` by switching to ``importlib.metadata.version("juniper-cascor")`` plus a real-module import.
- ``Verify Health Endpoint`` — passes; ``/v1/health/ready`` returns 200 with juniper_data dep status ``healthy``.

Final coverage TOTAL=92.38%; final unit-test summary 3115 passed, 15 skipped, 103 deselected. Ready to merge.

PR #187, opened 2026-05-03
feat(replay): emit weight payloads on sample-boundary events (Phase 6E CAN-015g, g-3) [stacked on #184]

pcalnon:
## Summary

Third PR of **CAN-015g** (Replay V2). **Stacked on #184** (g-2 — weight cache + extended response). Closes the cascor-side surface for V2 replay: synthetic \`epoch_end\` events now carry a base64-encoded weight payload on sample-boundary epochs so canopy can render decision-boundary playback as the scrubber moves.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) (juniper-ml).

## Implementation

- \`_ReplaySession.__init__\` builds an O(1) reverse lookup \`_epoch_to_sample = {epoch: sample_index}\` at session start when the cache is available. Empty for V1 snapshots.
- \`_emit_frame\` extended:
  - **V2 path** (\`cache.available\`): adds \`is_sample_boundary\` (bool). On a boundary epoch, additionally adds a \`weights\` block from \`weights_at(sample_index)\`, base64-encoded.
  - **V1 path** (cache absent / empty \`sample_indices\`): emits the unchanged pre-g-3 shape — no \`is_sample_boundary\`, no \`weights\`. Preserves backward compat for canopy clients that pre-date the V2 protocol.
- \`_encode_weight_payload\` walks the cache payload dict and builds the wire envelope per the plan §\"Synthetic epoch_end event (additive)\":

  \`\`\`json
  {
    \"sample_index\": 5,
    \"epoch\": 50,
    \"output_weights\": {\"dtype\": \"float32\", \"shape\": [4, 1], \"data\": \"<base64>\"},
    \"output_bias\":    {\"dtype\": \"float32\", \"shape\": [1],    \"data\": \"<base64>\"},
    \"hidden_units\": [
      {\"first_sample_index\": 1, \"activation\": \"tanh\",
       \"weights\": {\"dtype\": \"float32\", \"shape\": [3], \"data\": \"<base64>\"},
       \"bias\": 0.123}
    ]
  }
  \`\`\`

  Bias on hidden units stays a plain Python float (no base64 overhead for a single scalar).
- \`_encode_tensor\` is defensive: \`None\` / shape-less inputs return \`None\` so a sparse cache payload doesn't crash the playback thread.

## Tests (\`test_replay_weight_emission.py\`, 12 tests)

- **V1 backward-compat**: no \`is_sample_boundary\` field, no \`weights\`. V1 metric shape (epoch, replay, snapshot_id, etc.) unchanged.
- **Sample-boundary detection**: boundary epoch emits weights with correct \`sample_index\`; non-boundary epoch emits \`is_sample_boundary=False\` with no weights; first epoch (sample 0 at epoch 0) emits weights; out-of-range index skipped.
- **Tensor envelope**: \`output_weights\` / \`output_bias\` / hidden-units weights all round-trip through base64 → \`np.frombuffer\` → reshape, matching the source array exactly. Hidden-unit bias is a Python float, not an envelope. Units not yet added at the sample's index are excluded.
- **Empty \`weight_history\`** (\`sample_indices=[]\`) treated as V1 — no weight fields emitted.
- **Encoder defensive** against \`None\` inputs.

## Validation

- [x] 12 new + 19 g-2 cache + 27 lifecycle replay + serializer / snapshot tests = **197 passed** in JuniperCascor1 env (Python 3.13.13, GIL-enabled).
- [ ] Pre-existing failure in \`test_replay_control_range\` still present; unrelated to g-3 (verified in g-2 PR).

## Out of scope for g-3

- **Canopy-side decision-boundary consumption** → **g-4** (final cascor / canopy PR pair for CAN-015g).
- **Lifecycle-side population of \`network.weight_history\` during training** (V2 actually being recorded, not just played back). Currently \`weight_history\` is populated only when a g-1 snapshot is loaded for replay. Adding the live-capture path is a separate change — could be a CAN-015g extension or a follow-on item depending on UX priorities.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit clean
- [ ] Reviewer: confirm wire-format envelope shape matches the design's spec exactly so g-4 (canopy) can implement against this PR's contract without surprises

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -303,6 +303,12 @@ class _ReplaySession:
         # eagerly even in the V1 case so consumers don't have to
         # branch on ``is None``.
         self.weight_cache = _WeightCache(weight_history, byte_budget=weight_cache_budget_bytes if weight_cache_budget_bytes is not None else _WeightCache.DEFAULT_BUDGET_BYTES)
+        # CAN-015g (g-3): O(1) reverse lookup from time_index (0-based
+        # epoch in the metric arrays) → sample_index (0-based into
+        # sample_indices). Populated only when the cache is available;
+        # the emitter checks membership to decide whether to attach a
+        # ``weights`` block to the synthetic event.
+        self._epoch_to_sample: Dict[int, int] = ({int(epoch): i for i, epoch in enumerate(self.weight_cache.sample_indices)} if self.weight_cache.available else {})
         # Playback state — guarded by the lock for cross-thread reads.
         self._lock = threading.Lock()
         self.time_index: int = 0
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #188, opened 2026-05-03
metrics(r5-4-pre): cascor training counters + train-step histogram + worker→Prometheus bridge

pcalnon:
## Summary

Closes the three instrumentation gaps that R5.1's SLO catalog
([juniper-deploy#48](https://github.com/pcalnon/juniper-deploy/pull/48))
flagged as unbinable. R5.4 (burn-rate alerts in juniper-deploy) is
gated on this PR landing.

### Gap 1 — `juniper_cascor_training_sessions_completed_total{status}` (counter)

Closed-set `status` label values per R1.1 cardinality discipline:
`success` / `failure` / `cancelled`. Bumped exactly once per terminal
FSM transition by the lifecycle manager's `monitored_fit` wrapper:
- `mark_completed` path → `success`
- `Command.STOP` mid-fit path → `cancelled`
- exception path → `failure`

**Binds SLO 3.3** ([juniper-deploy#48](https://github.com/pcalnon/juniper-deploy/pull/48)
SLO_CATALOG §3.3): training-session success ratio.

```promql
sum(rate(juniper_cascor_training_sessions_completed_total{status="success"}[5m]))
  / sum(rate(juniper_cascor_training_sessions_completed_total[5m]))
```

### Gap 2 — `juniper_cascor_training_step_duration_seconds{phase}` (histogram)

Buckets `(0.05, 0.1, 0.5, 1.0, 2.0, 5.0, 10.0, 30.0, +inf)` target SLO
3.4 (p95 < 5 s). Emitted at the **api-lifecycle layer** via the existing
per-epoch output callback using `time.perf_counter` deltas. The `phase`
label is currently always `"output"` because only the output-phase
epoch callback is instrumented; candidate / input phase emission is on
the R5.4 follow-up queue.

**Born SLO-aligned** — HELP string carries the SLO 3.4 callout and does
NOT carry the R4.1 "tentative pending R5.1" suffix. Per-fit timer is
reset by `monitored_fit` so observations don't leak across fit()
invocations.

Histogram-buckets rationale doc updated:
- §1.1 — train-step boundary discussion (api-lifecycle vs. neural-net layer)
- §6 — per-boundary rationale for the new metric
- §7 / §8 — renumbered (was §6 / §7)

**Binds SLO 3.4** ([juniper-deploy#48](https://github.com/pcalnon/juniper-deploy/pull/48)
SLO_CATALOG §3.4).

### Gap 3 — `api.workers.metrics.WorkerRegistryCollector` (bridge)

[...]

PR #191, opened 2026-05-03
feat(training): capture per-sample weight history during training (Phase 6E CAN-015g, g-6)

pcalnon:
## Summary

Sixth PR of **CAN-015g** (Replay V2). Closes the V2 protocol's **write-side gap**: production training runs now actually populate \`network.weight_history\`, so saved snapshots carry the per-sample tensors that g-1's serializer persists, g-2's cache loads, and g-3's emitter emits.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) §\"Live capture during training (g-6)\" (added in juniper-ml #188).

## Discovery context

While implementing g-1/g-2/g-3 it surfaced that the cascor side could **play back** a \`weight_history\` but nothing in the training loop **populated** it — the protocol worked end-to-end against ad-hoc test fixtures only. g-6 closes the gap.

## Implementation

- **New \`_WeightHistoryRecorder\` class** in \`lifecycle/manager.py\`. Three triggers:
  - **Periodic**: every Nth epoch via \`training_monitor.on_epoch_end\`. \`N\` from \`config.weight_history_sampling_interval\` (default 50; \`N=1\` captures every epoch — Option A; \`N=0\` disables periodic — Option D).
  - **Cascade-grow**: every \`training_monitor.on_cascade_add\` event. Marked decimation-exempt.
  - **Terminal**: \`capture_terminal()\` from \`monitored_fit\`'s post-fit block (covers cancelled-by-stop and clean-completion paths).
- **Idempotent dedupe by epoch** — single epoch firing both Nth-epoch and cascade-add triggers records one sample (cascade-add wins).
- **Memory ceiling** via \`config.weight_history_max_samples\` (default 1000). On overflow, \`_decimate\` halves inter-cascade density while always retaining cascade-add samples; recorded \`sampling_interval\` doubles to reflect the on-disk gap.
- **Hidden-unit slicing matches the g-2 cache convention**: \`first_sample_index\` is a 0-based index into \`sample_indices\`, NOT an epoch number.
- **Tensor capture** uses \`.detach().cpu().numpy().copy()\` — drops autograd retention, produces buffers the optimizer can't mutate later.
- **Best-effort**: capture exceptions caught + logged but never crash the training thread.

## Wiring

- \`lifecycle/manager.py\`: new \`self._weight_history_recorder\` attribute, lazy attach via \`_attach_weight_history_recorder()\` from \`start_training\` just before submitting the future. Re-attach picks up runtime PATCH \`/v1/training/params\` changes.
- Terminal capture wired into \`monitored_fit\`'s post-fit block.

## Config

Two new tunables on \`CascadeCorrelationConfig\`:

| Field | Default | Purpose |
|---|---|---|
| \`weight_history_sampling_interval\` | 50 | N for the every-Nth-epoch trigger. \`1\` = every epoch (Option A). \`0\` = disable periodic (Option D — cascade-add only). |
| \`weight_history_max_samples\` | 1000 | Soft cap before decimation kicks in. \`0\` = unbounded. |

Both runtime-mutable via the existing PATCH \`/v1/training/params\` flow.

## Tests (\`test_weight_history_recorder.py\`, 17 tests)

- **Initialization**: pre-g-6 networks pick up V2 capture; existing \`weight_history\` preserved on re-attach (only \`sampling_interval\` refreshed); \`register()\` is idempotent.
- **Periodic trigger**: captures every Nth epoch; \`sampling_interval=0\` disables it; captured sample is an independent copy.
- **Cascade-add trigger**: captures at monitor's \`current_epoch\`; dedupes when same epoch fires both periodic + cascade-add.
- **Terminal capture**: marks sample as decimation-exempt.
- **Decimation**: cascade samples retained; \`max_samples=0\` disables; \`sampling_interval\` recorded value doubles after a decimation pass.
- **Hidden-unit slicing**: \`first_sample_index\` is a sample-list index; per-unit bias recorded as Python float.
- **Robustness**: capture exception in tensor copy doesn't crash the monitor; \`capture_terminal\` without monitor state is a no-op.

## Validation

- [x] 17 new + 165 existing lifecycle/serializer tests = **182 passed** in JuniperCascor1 (Python 3.13.13).

## Out of scope for g-6

- **Wall-clock benchmark** (≤ 5% overhead vs. baseline) — captured in the plan's risk register; will land as a perf regression test once the fast-slow mode test harness gets a benchmark fixture for it.
- **Decimation chaos test** (100k-epoch synthetic loop) — captured in the plan; deferred until the perf-test harness is wired.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit clean
- [ ] Reviewer: confirm the trigger mix (every Nth + cascade + terminal) matches the plan's intent. The dedupe-by-epoch behavior is testable via \`test_cascade_add_dedup_with_periodic\`.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -61,6 +61,399 @@ def _write_activation_function_name(network: Any, value: str) -> None:
     network._init_activation_function()
 
 
+class _WeightHistoryRecorder:
+    """CAN-015g (Phase 6E follow-on, g-6): training-loop weight history capture.
+
+    Populates ``network.weight_history`` during a training run so the
+    serializer's V2 path (g-1) actually has data to persist when the
+    user calls ``POST /v1/snapshots``. Without this, V2 replay only
+    works against ad-hoc test fixtures — production training would
+    still produce V1-only snapshots.
+
+    Plan: ``notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md`` §"Live
+    capture during training (g-6)".
+
+    Three trigger points:
+      1. **Every Nth epoch** — registered as a callback on
+         ``training_monitor.on_epoch_end``. ``N`` = network's
+         ``config.weight_history_sampling_interval`` (default 50;
+         set to 1 for every-epoch capture; set to 0 to disable the
+         periodic trigger and rely on cascade-add only).
+      2. **Cascade-grow events** — registered as a callback on
+         ``training_monitor.on_cascade_add``. Always captures
+         regardless of the periodic interval since these are the
+         narrative anchors per the parent design.
+      3. **Terminal capture** — call ``capture_terminal()`` from the
+         lifecycle's training-completion path so the last sample
+         reflects the truly-final weights even when training stops
+         mid-interval.
+
+    Memory ceiling: ``config.weight_history_max_samples`` (default
+    1000) is a soft cap. On overflow the recorder decimates
+    inter-cascade samples by 2× while always retaining cascade-add
+    samples (they're the visually-meaningful moments).
+
+    Idempotency: a single epoch can fire both Nth-epoch and
+    cascade-add triggers. The recorder dedupes by epoch number — a
+    sample for a given epoch is recorded at most once, with the
+    cascade-add capture winning if both fire (it has the latest
+    post-grow state).
+
+    Thread safety: capture runs in the training thread (the same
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #193, opened 2026-05-03
fix(observability): make Prometheus metric registration idempotent (P-7)

pcalnon:
Cascor main red since PR #188 with `Duplicated timeseries in CollectorRegistry: {'juniper_cascor_training_sessions_active'}` on 4 unit tests. The lazy-singleton pattern in `src/api/observability.py` doesn't survive tests that clear the cache without unregistering — the metric *name* stays in REGISTRY but the Python reference is lost. Fix: wrap each metric ctor in a `_register_or_reuse` helper that on `Duplicated timeseries` looks the orphan up in `REGISTRY._collector_to_names`, unregisters it, and retries. Behavior in production (no orphans) is unchanged. Refs juniper-ml notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md (P-7).

### DIFF
--- a/src/api/observability.py
+++ b/src/api/observability.py
@@ -159,6 +159,40 @@ _TRAINING_STEP_DURATION_BUCKETS: tuple = (
 )
 
 
+def _register_or_reuse(cls, name: str, *args, **kwargs):
+    """Construct a Prometheus collector idempotently against the default registry.
+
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #197, opened 2026-05-03
test(replay): align range-dict assertions with PR #195's contract (P-8)

pcalnon:
PR #195 reverted the production code to keep `range == {start, end}` and surface `time_index` as a top-level state_summary field, but didn't update the two tests that asserted the old three-key shape. Cascor main has been red since #195 merged.

This PR updates both tests to the new contract:
- `state["range"] == {"start", "end"}` (no `time_index`)
- `state["time_index"]` asserted separately as a top-level field

No production-code changes — assertion shape only.

Refs juniper-ml notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md (P-8 follow-up to P-2 cluster B / PR #195).


pcalnon (commented):
approved

PR #198, opened 2026-05-03
refactor(network): extract _install_hidden_unit helper (no behaviour change) (Phase 6E CAN-015h-0 prep)

pcalnon:
## Summary

Refactor prerequisite for **CAN-015h**. The cascade-grow paths (\`add_unit\` and \`add_units_as_layer\`) and the future \`POST /v1/network/hidden-units\` endpoint (h-2) all share the same conceptual operation: append a unit dict to \`hidden_units\`, append metadata to \`history[\"hidden_units_added\"]\`, and rebuild \`output_weights\` to widen the output layer. Without a shared helper, h-2 would either duplicate the code or import private implementation details from the training loop. This PR extracts the shared logic into two narrow helpers with clear contracts.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) §\"PR breakdown — CAN-015h\" h-0 (juniper-ml).

## New helpers

- **\`_install_hidden_unit(weights, bias, activation_fn, correlation=0.0)\`** — appends one unit (clone+detached) to \`hidden_units\` and one metadata record to \`history[\"hidden_units_added\"]\`. Returns the new unit's 0-based index. Does **NOT** touch \`output_weights\` so callers can batch resizes (\`add_units_as_layer\` needs this for N-unit layer growth).
- **\`_resize_output_layer_for_new_units(num_added, prev_input_size)\`** — widens \`output_weights\` by \`num_added\` rows per \`init_output_weights\` strategy (\"zero\" or \"random×0.1\"), copies old weights into the \`[:prev_input_size, :]\` prefix slice, enables gradient tracking on the new weight matrix, and clones+detaches \`output_bias\` (preserving the pre-refactor side-effect that stripped its \`requires_grad\`). No-op when \`num_added == 0\`.

## Refactored callers

- **\`add_unit\`** — calls \`_install_hidden_unit\` then \`_resize_output_layer_for_new_units(1, ...)\`. RNG-consumption point identical to the pre-refactor inline path.
- **\`add_units_as_layer\`** — calls \`_install_hidden_unit\` per successful candidate, then a single batched \`_resize_output_layer_for_new_units(added_count, ...)\`.

## RNG sensitivity

The random-init path consumes one \`torch.randn\` call per resize. The refactor preserves the call-site location so deterministic-seed tests downstream see identical tensor values. The bit-identity test \`test_random_init_path_consumes_one_randn_call\` asserts this by running the same fixed-seed \`add_unit\` twice and comparing tensors.

## Tests (\`test_install_hidden_unit_helper.py\`, 13 tests)

- **\`_install_hidden_unit\`**: appends to \`hidden_units\`; appends to \`history\` with correct \`unit_index\`; clones inputs (no aliasing on source-tensor mutation); returns monotonic indices across sequential calls.
- **\`_resize_output_layer_for_new_units\`**: widens to expected shape (zero init); preserves old weights in prefix slice; zero-init fills new rows with zeros; enables grad on \`output_weights\`; **strips grad from \`output_bias\`** (preserves pre-refactor side-effect); no-op for \`num_added=0\`.
- **Bit-identity**: random-init \`add_unit\` produces deterministic tensors under fixed seed; refactored \`add_unit\` yields expected structure (1 unit added, 1 row added to \`output_weights\`, 1 history record); two sequential adds account correctly.

## Validation

- [x] 13 new + 426 existing cascade_correlation tests = **439 passed** in JuniperCascor1 (Python 3.13.13).

## Out of scope

- The PATCH \`/weights\`, POST \`/hidden-units\`, DELETE \`/hidden-units/{idx}\` endpoints — those are **h-1**, **h-2**, **h-3** and consume the helpers introduced here.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit clean
- [ ] Reviewer: confirm the bit-identity test pattern (deterministic re-run + tensor equality) is enough to catch RNG-order regressions, or flag if you want a captured-against-pre-refactor reference instead

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -3428,6 +3428,96 @@ class CascadeCorrelationNetwork:
 
     #################################################################################################################################################################################################
     # Public Method to add a new hidden unit based on the correlation
+    def _install_hidden_unit(
+        self,
+        weights: torch.Tensor,
+        bias: torch.Tensor,
+        activation_fn,
+        correlation: float = 0.0,
+    ) -> int:
+        """CAN-015h-0: shared helper that appends a hidden unit to the network.
+
+        Used by both the cascade-grow path (``add_unit`` /
+        ``add_units_as_layer``) and the future
+        ``POST /v1/network/hidden-units`` endpoint (h-2). Keeping
+        the unit-installation logic in one place prevents the two
+        callers from drifting apart on dict shape, history-record
+        layout, or unit-index accounting.
+
+        The helper is **install only** — it does NOT resize
+        ``self.output_weights``. Call
+        ``_resize_output_layer_for_new_units`` separately. The
+        two-step split lets ``add_units_as_layer`` batch one resize
+        for N units instead of N resizes.
+
+        Args:
+            weights: Cloned + detached weight tensor for the new unit.
+            bias: Cloned + detached bias tensor for the new unit.
+            activation_fn: Callable applied to the unit's pre-activation.
+            correlation: Recorded in history. ``0.0`` is a sane
+                default for manual inserts where correlation is
+                undefined.
+
+        Returns:
+            int: 0-based index of the appended unit (i.e.
+            ``len(self.hidden_units) - 1`` after the append).
+        """
+        new_unit = {
+            "weights": weights.clone().detach(),
+            "bias": bias.clone().detach(),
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #199, opened 2026-05-03
feat(network): PATCH /v1/network/weights with shape + NaN validation (Phase 6E CAN-015h-1)

pcalnon:
## Summary

First mutation endpoint of **CAN-015h**. Wires the post-restore \`Investigating\` FSM state to a surgical-rewrite endpoint that canopy's network-editor panel (h-5) consumes.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) §\"Endpoint design / 1. PATCH /v1/network/weights\" (juniper-ml).

## API surface

\`\`\`http
PATCH /v1/network/weights
Content-Type: application/json

{
  \"target\": \"output\" | \"hidden_unit\",
  \"field\": \"weights\" | \"bias\",
  \"values\": [[...], [...]],
  \"hidden_unit_index\": 3,        // required iff target == \"hidden_unit\"
  \"dtype\": \"float32\"             // optional; default float32
}
\`\`\`

Status codes:

- **200** — patch applied
- **400** — shape mismatch / unknown target or field
- **404** — no network created / hidden_unit_index out of range
- **409** — FSM not Investigating
- **422** — NaN/Inf or untensorable values

## Lifecycle (\`api/lifecycle/manager.py\`)

- **\`patch_weights\` returns a sentinel-status dict** so the route layer can map errors to HTTP statuses cleanly without HTTPException leaking into the lifecycle layer.
- **FSM gate**: rejects from any state other than \`INVESTIGATING\` (entered via \`/restore\`).
- **Validation**: exact-shape match, NaN/Inf rejection (including post-cast Inf detection), ragged-list catch via tensor-coerce failure path.
- **Optimizer-state zero-out**: runs **before** the parameter reassignment because the optimizer's state dict is keyed by tensor identity. Skips 0-d scalar tensors (Adam's \`step\` counter is stored as a scalar in newer PyTorch); zeros n-d running-statistic buffers (\`exp_avg\`, \`exp_avg_sq\`).
- **Best-effort**: missing optimizer / missing state / non-dict state are all silent no-ops.

## Tests (\`test_patch_weights.py\`, 27 tests)

- **Validation**: no-network, unknown target/field, missing \`hidden_unit_index\`, out-of-range index, shape mismatch, NaN, Inf, unparsable values.
- **Output-layer patches**: weights rewrite, \`requires_grad\` preserved, bias rewrite.
- **Hidden-unit patches**: weights rewrite, bias rewrite, shape mismatch.
- **Optimizer-state zero-out**: output_weights case + hidden-unit case + step counter preserved + missing optimizer is no-op + missing state for param is no-op.

## Validation

- [x] 27 new + 138 existing lifecycle/install_hidden_unit tests = **165 passed** in JuniperCascor1.
- 122 pre-existing failures in \`test_api_security\` / \`test_websocket_manager\` / \`test_control_stream_*\` — verified identical pre/post change.

## Out of scope

- **h-2**: POST \`/v1/network/hidden-units\` with cascade rebuild (next PR).
- **h-3**: DELETE \`/v1/network/hidden-units/{idx}\` with cascade rebuild.
- **h-4 through h-6**: canopy adapter + Network Editor panel.

## Test plan

- [x] Unit tests pass
- [x] Pre-commit clean
- [ ] Reviewer: confirm the sentinel-status pattern feels right vs. raising exceptions inside the lifecycle. The choice keeps the lifecycle layer HTTP-agnostic so it stays test-friendly without FastAPI fixtures.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -2224,6 +2224,175 @@ class TrainingLifecycleManager:
             return {}
 
     # ------------------------------------------------------------------
+    # CAN-015h-1: PATCH /v1/network/weights — surgical weight edit
+    # ------------------------------------------------------------------
+
+    # Sentinel values returned to the route layer so the route can map
+    # them to the right HTTP status. Plain string codes keep the
+    # lifecycle-↔-route boundary serializable for tests without
+    # requiring HTTPException to escape into the lifecycle layer.
+    _PATCH_OK: str = "ok"
+    _PATCH_FSM_REJECTED: str = "fsm_rejected"
+    _PATCH_NO_NETWORK: str = "no_network"
+    _PATCH_BAD_TARGET: str = "bad_target"
+    _PATCH_SHAPE_MISMATCH: str = "shape_mismatch"
+    _PATCH_NAN_INF: str = "nan_inf"
+    _PATCH_HIDDEN_UNIT_OUT_OF_RANGE: str = "hidden_unit_out_of_range"
+
+    def patch_weights(
+        self,
+        target: str,
+        field: str,
+        values: Any,
+        hidden_unit_index: Optional[int] = None,
+        dtype: str = "float32",
+    ) -> Dict[str, Any]:
+        """CAN-015h-1: surgically rewrite a single parameter group.
+
+        Returns a dict ``{"status": <code>, "detail": <str>}`` where
+        ``status`` is one of the ``_PATCH_*`` sentinels. The route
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #200, opened 2026-05-03
feat(network): POST /v1/network/hidden-units with cascade rebuild (Phase 6E CAN-015h-2) [stacked on #199]

pcalnon:
## Summary

Second mutation endpoint of **CAN-015h**. **Stacked on #199** (h-1). Appends a fresh hidden unit at the cascade tail while the FSM is \`Investigating\`; reuses h-0's \`_install_hidden_unit\` and \`_resize_output_layer_for_new_units\` helpers so the manual-add path can never drift apart from the training-loop cascade-grow path.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) §\"Endpoint design / 2. POST /v1/network/hidden-units\" (juniper-ml).

## V1 limitation

\`position=\"tail\"\` only. Insert-at-position is documented in the plan but **deferred** here — the connectivity-rewrite semantics are user-surprising and worth shipping with explicit canopy UI text. A future PR will add \`position: int\` once that text exists. The Pydantic body type forces this at the boundary (\`position: Literal[\"tail\"]\`).

## API surface

\`\`\`http
POST /v1/network/hidden-units
Content-Type: application/json

{
  \"weights\": [...],                   // length = input_size + num_existing_hidden_units
  \"bias\": 0.0,
  \"activation\": \"Tanh\",
  \"position\": \"tail\"                 // V1: tail-only
}
\`\`\`

Status codes: **200** (appended) / **400** (shape mismatch) / **404** (no network) / **409** (FSM not Investigating, or at \`max_hidden_units\` cap) / **422** (NaN/Inf, unknown activation).

## Implementation highlights

- **Forced zero-init of the new output column** regardless of \`init_output_weights\` config. Rationale: the user hasn't trained against this unit, so a non-zero output contribution would be arbitrary. Save / restore \`init_output_weights\` so the network's persistent config stays whatever the user originally set.
- **Optimizer dropped to None** after append. Adam's state tensors no longer match the new parameter shape; next training pass rebuilds from scratch.
- **History records \`correlation=0.0\`** as a sentinel for manual inserts (training-loop adds carry the candidate's actual correlation; downstream consumers can distinguish if needed).
- **\`max_hidden_units\` cap surfaces 409** rather than raising \`ValidationError\` (training-loop contract). The manual endpoint should give canopy a friendly error code.

## Tests (\`test_add_hidden_unit_manual.py\`, 21 tests)

- **Validation**: no-network, at-cap, unknown activation, NaN weights, Inf bias, shape mismatch, 2-D weights rejected.
- **Successful append**: hidden_units grows by one, output_weights widened by one row, **new column is zero** (regardless of the original init config), original \`init_output_weights\` preserved, history records \`correlation=0.0\`, optimizer dropped, two sequential appends account correctly.

## Validation

- [x] 21 new + h-1's 27 + 138 sibling lifecycle tests = **186 passed** in JuniperCascor1.

## Out of scope

- **h-3**: DELETE \`/v1/network/hidden-units/{idx}\` (next PR — final cascor-side endpoint).
- **h-4 through h-6**: canopy adapter + Network Editor panel.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -2393,6 +2393,128 @@ class TrainingLifecycleManager:
             param_state[key] = torch.zeros_like(val)
 
     # ------------------------------------------------------------------
+    # CAN-015h-2: POST /v1/network/hidden-units — manual unit append
+    # ------------------------------------------------------------------
+
+    _ADD_OK: str = "ok"
+    _ADD_FSM_REJECTED: str = "fsm_rejected"
+    _ADD_NO_NETWORK: str = "no_network"
+    _ADD_AT_CAP: str = "at_cap"
+    _ADD_BAD_ACTIVATION: str = "bad_activation"
+    _ADD_BAD_SHAPE: str = "bad_shape"
+    _ADD_NAN_INF: str = "nan_inf"
+
+    def add_hidden_unit_manual(
+        self,
+        weights: Any,
+        bias: float = 0.0,
+        activation: str = "Tanh",
+    ) -> Dict[str, Any]:
+        """CAN-015h-2: append a hidden unit at the cascade tail.
+
+        Companion to ``add_unit`` (training-loop cascade-grow), but
+        invoked manually from the canopy editor while the FSM is
+        Investigating. Reuses the h-0 helpers
+        (``_install_hidden_unit`` + ``_resize_output_layer_for_new_units``)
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #201, opened 2026-05-03
feat(network): DELETE /v1/network/hidden-units/{idx} with cascade rebuild (Phase 6E CAN-015h-3) [stacked on #200]

pcalnon:
## Summary

Third (final cascor-side) mutation endpoint of **CAN-015h**. **Stacked on #200** (h-2). Removes the hidden unit at \`idx\` with surgical cascade rebuild.

Plan reference: [\`notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`](https://github.com/pcalnon/juniper-ml/blob/main/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md) §\"Endpoint design / 3. DELETE /v1/network/hidden-units/{idx}\" (juniper-ml).

## The cascade-rebuild surgery (the point of this PR)

Each hidden unit \`j\` has a weight vector of length \`input_size + j\`. Unit \`j\`'s weight at position \`input_size + i\` (for \`i < j\`) is the connection from unit \`i\`'s output. When we remove unit \`idx\`, every subsequent unit's connection to the deleted unit's output must be dropped:

1. Unit \`j > idx\` had its connection at position \`input_size + idx\`. **Drop that single weight from \`j\`'s weight vector.**
2. After the surgery, what was unit \`j\` is now unit \`j - 1\` with a weight vector of length \`input_size + (j - 1)\` — consistent with its new cascade position.
3. The corresponding column \`input_size + idx\` of \`output_weights\` is removed.

The forward-pass shape invariant holds post-delete — verified by \`test_post_delete_forward_pass_succeeds\`.

### Approach choice

This is more invasive than h-2's tail-only append. The plan's intent was clearly \"middle deletion supported.\" Two alternatives were rejected:

- **Catastrophic rebuild** (drop all subsequent units): too destructive — user mental model is \"remove this one unit,\" not \"destroy half the network.\"
- **Tail-only delete** (V1 restriction matching h-2's tail-only add): conservative but loses the most useful UX case — deleting a known-bad middle unit.

## API surface

\`\`\`http
DELETE /v1/network/hidden-units/{idx}
\`\`\`

Status codes: **200** (removed) / **404** (no network OR \`idx\` out of range, per the design plan) / **409** (FSM not Investigating).

## Tests (\`test_remove_hidden_unit_manual.py\`, 19 tests)

- **Validation**: no-network, negative idx, \`idx==count\`, empty network.
- **Tail removal**: structurally identical to truncating the last row; remaining units' weights unchanged.
- **Middle removal**: cascade-rebuild surgery covers removing unit 0 (everything shifts), removing unit 1 (middle of 3), and the **forward-pass shape invariant** post-delete.
- **Output weights surgery**: column dropped (verified by stamping a recognizable pattern and checking which entries survive); optimizer dropped; \`requires_grad\` preserved.

## Validation

- [x] 19 new + h-2's 21 + h-1's 27 + 138 sibling lifecycle tests = **205 passed** in JuniperCascor1.

## CAN-015h cascor side complete

After h-1 / h-2 / h-3 land, the cascor backend exposes the full mutation surface:

- \`PATCH /v1/network/weights\` (h-1)
- \`POST /v1/network/hidden-units\` (h-2)
- \`DELETE /v1/network/hidden-units/{idx}\` (h-3)

All FSM-gated to \`Investigating\`. Canopy editor panel + adapter (h-4 through h-6) build on top.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -2515,6 +2515,90 @@ class TrainingLifecycleManager:
         }
 
     # ------------------------------------------------------------------
+    # CAN-015h-3: DELETE /v1/network/hidden-units/{idx} — manual remove
+    # ------------------------------------------------------------------
+
+    _REMOVE_OK: str = "ok"
+    _REMOVE_FSM_REJECTED: str = "fsm_rejected"
+    _REMOVE_NO_NETWORK: str = "no_network"
+    _REMOVE_OUT_OF_RANGE: str = "out_of_range"
+
+    def remove_hidden_unit_manual(self, idx: int) -> Dict[str, Any]:
+        """CAN-015h-3: remove the hidden unit at ``idx``.
+
+        Cascade-rebuild semantics:
+
+        - Unit ``idx`` is removed.
+        - Each subsequent unit ``j > idx`` had a weight at position
+          ``input_size + idx`` that referenced the removed unit's
+          output. That single weight is dropped from each subsequent
+          unit's weight vector. After the surgery, what was unit
+          ``j`` is now unit ``j - 1`` with weight vector length
+          ``input_size + (j - 1)`` — consistent with its new
+          cascade position.
+        - The corresponding column ``input_size + idx`` of
+          ``output_weights`` is removed; ``output_weights`` reshapes
+          to ``[in + num_units - 1, out]``.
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #181, opened 2026-05-03
test(v38): retag flaky grow_network tests as integration (Option E fallback)

pcalnon:
## Summary

Closes V38a/V38c per the planning doc's Option E fallback (`juniper-ml/notes/V38_GROW_NETWORK_INVESTIGATION_PLAN_2026-05-02.md`).

The 4 flaky tests that asserted exact `hidden_units` counts after `grow_network()`/`fit()` calls are retagged `@pytest.mark.integration` so the unit-test gate stops blocking on them; they continue to run in the integration tier.

## Diagnostic findings (Phase A.2)

CI run on `debug/v38-grow-network-diagnostic` HEAD `d0684fd` ([25266874480](https://github.com/pcalnon/juniper-cascor/actions/runs/25266874480)) showed all 4 V38a/V38c tests **passing on every leg** when `CASCOR_GROW_DEBUG=1`-gated `print(..., flush=True)` calls were injected at 4 checkpoints in `train_candidates` / `_execute_candidate_training` / `_process_training_results`. Per-result diagnostic confirmed each candidate trained healthily (correlation ~0.52-0.55, success=True, error_message=None) — **NOT** RC-3 (silent exception). The test failure is a **classic Heisenbug**: I/O-sync timing of the `flush=True` calls consistently dodges a race window in the candidate-training dispatcher. **Confirmed RC-4 (multiprocessing fragility)** per the planning doc.

The underlying race lives somewhere in `_execute_candidate_training` ↔ `_task_distributor` boundary and remains an open tracked issue. Investigation requires a torch-capable repro environment (the conda envs on the workstation have a `torch._C` shadow issue that blocks local repro).

## Changes

`src/tests/unit/test_cascade_correlation_coverage.py`:
- `TestNetworkGrowth.test_grow_network_adds_hidden_unit`: `@pytest.mark.unit` → `@pytest.mark.integration`.

`src/tests/unit/test_fast_fit.py`:
- Removed class-level `@pytest.mark.unit`. Three count-asserting tests retagged `@pytest.mark.integration`:
  - `test_fit_executes_full_training_path`
  - `test_fit_output_weights_grow_after_unit_addition`
  - `test_fit_multiple_iterations`
- Other two tests (`test_fit_with_validation_data`, `test_fit_history_tracks_accuracy`) keep `@pytest.mark.unit` explicitly — they exercise `fit()` but assert only history-key presence and pass deterministically.

[...]

PR #203, opened 2026-05-03
fix(parallelism): close RC-4 candidate-training race + ring-buffer instrumentation (P-1)

pcalnon:
Closes **P-1** — the RC-4 multiprocessing race in cascor's candidate-training distributor that V38 Phase A.2 narrowed to a heisenbug at the `_task_distributor.distribute_and_collect` IPC boundary.

## Root cause

`_collect_worker_results` polled `result_queue.qsize()` to break out of its wait loop early. `multiprocessing.Queue.put()` is non-atomic with `qsize()`: the worker's `put` returns when the pickle lands in the queue's in-process `_buffer` deque, but `qsize()` only increments after a daemon feeder thread writes to the underlying `Pipe` and bumps the queue's semaphore. Under sub-second per-candidate budgets the parent's qsize check raced ahead of in-flight feeder writes, exited early, and `_collect_training_results` then timed out on results whose `_sem` hadn't been bumped — yielding silent `best_candidate=None`.

`print(flush=True)` masked the race during V38 because stdout-flush + disk-sync gave the feeder threads enough wall-clock to land their `_sem` bumps.

## Fix (H1 from [the plan](https://github.com/pcalnon/juniper-ml/blob/main/notes/P1_RC4_INVESTIGATION_PLAN_2026-05-03.md))

Drop the qsize-poll wait loop entirely. `_collect_training_results`'s `get(timeout=...)` is already correctly synchronized with the queue's semaphore and feeder thread. Forcing every collection path through that single, correctly-synchronized call closes the race in both fast and slow regimes.

The four V38a/V38c tests retagged `@pytest.mark.integration` (cascor #181) are flipped back to `@pytest.mark.unit`:

- `test_fast_fit.py::TestFastFit::test_fit_executes_full_training_path`
- `test_fast_fit.py::TestFastFit::test_fit_output_weights_grow_after_unit_addition`
- `test_fast_fit.py::TestFastFit::test_fit_multiple_iterations`
- `test_cascade_correlation_coverage.py::TestNetworkGrowth::test_grow_network_adds_hidden_unit`

## Ring-buffer instrumentation

`src/parallelism/rc4_ring_buffer.py` — gated by `CASCOR_RC4_RING_BUFFER=1`, no-op otherwise. Captures parent + worker events without the I/O sync that masked the race during V38. `pytest_runtest_makereport` in `conftest.py` dumps the buffer chronologically on any test failure when the flag is set, so future regressions surface their signature without a re-bisect.

## Verification

- 50× pytest-repeat over the four formerly-flaky tests under JuniperCascor1 (Python 3.13 + torch 2.11.0+cu130): all pass.
- Full unit suite: 3 192 passed, 15 skipped, 99 deselected (vs prior 3 188 — +4 from the retag).
- Pre-commit clean.

## Test plan

- [ ] CI's Unit Tests + Coverage matrix is green (Python 3.12/3.13/3.14 × ubuntu/macOS) **with the four formerly-retagged tests now in the unit tier**.
- [ ] If any of those four (or any other test) trips a residual race in CI, the ring-buffer dump in the captured stdout pinpoints it.
- [ ] No regression elsewhere.

Refs:
- [`V38_GROW_NETWORK_INVESTIGATION_PLAN_2026-05-02.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/V38_GROW_NETWORK_INVESTIGATION_PLAN_2026-05-02.md) §7 (Phase A.2 narrowing).
- [`POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/POST_V38_OPEN_ISSUES_PLAN_2026-05-03.md) (P-1).
- [`P1_RC4_INVESTIGATION_PLAN_2026-05-03.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/P1_RC4_INVESTIGATION_PLAN_2026-05-03.md) §4 (fix-shape ranking).

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -132,6 +132,11 @@ from cascor_constants.constants import (  # TODO: Commented out for F401 complia
 from cascor_plotter.cascor_plotter import CascadeCorrelationPlotter
 from log_config.log_config import LogConfig
 from log_config.logger.logger import Logger
+
+# P-1 / RC-4 ring buffer (gated by CASCOR_RC4_RING_BUFFER=1; no-op
+# otherwise). See src/parallelism/rc4_ring_buffer.py and
+# notes/P1_RC4_INVESTIGATION_PLAN_2026-05-03.md (juniper-ml).
+from parallelism import rc4_ring_buffer as _rc4
 from parallelism.task_distributor import TaskDistributor
 from utils.utils import display_progress
 
@@ -2002,6 +2007,7 @@ class CascadeCorrelationNetwork:
             try:
                 stale_result = result_queue.get_nowait()
                 drained_count += 1
+                _rc4.emit("parent.drain.stale", candidate_id=getattr(stale_result, "candidate_id", None), round_id=getattr(stale_result, "round_id", None))
                 self.logger.warning(f"CascadeCorrelationNetwork: _drain_stale_results: " f"Drained stale result from previous round: " f"candidate_id={getattr(stale_result, 'candidate_id', '?')}, " f"correlation={getattr(stale_result, 'correlation', '?')}")
             except _QueueEmpty:
                 break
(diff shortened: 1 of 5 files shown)


pcalnon (commented):
approved

PR #205, opened 2026-05-03
fix(tests): session-end cleanup — reap orphans + flush stdout before os._exit

pcalnon:
## Summary

Repairs three related session-end behaviors in ``src/tests/conftest.py``:

- **Issue 3**: ``multiprocessing`` forkserver / worker / resource-tracker children left behind when pytest exits, which accumulated to ~12 GB RSS across worktrees on 2026-05-03 and triggered an OOM feedback loop on subsequent pytest runs.
- **Issue 2**: pytest's ``X passed in Ys`` summary line silently dropped when ``pytest_unconfigure`` takes the ``os._exit(0)`` branch, because ``os._exit`` skips the implicit stdout flush that normal Python exit performs.

See ``project_cascor_orphan_forkserver_workers_2026-05-03`` and ``project_cascor_pytest_summary_truncation_2026-05-03`` in juniper-ml auto-memory for the full failure patterns.

### Step 2 — per-test cleanup (Issue 3)

New autouse function-scope ``_reap_test_spawned_children`` records the ``multiprocessing.active_children()`` set at test setup and, after the test, terminates only the children that appeared during that test. Pre-existing children are left alone so performance tests that intentionally reuse a session-scoped pool keep working.

### Step 3 — session-end safety net (Issue 3)

``pytest_unconfigure(trylast=True)`` now calls ``_reap_multiprocessing_children`` before ``os._exit(0)``:

  1. Walk ``multiprocessing.active_children()``, terminate, join with deadline, SIGKILL stragglers.
  2. Explicitly stop ``multiprocessing.forkserver._forkserver`` and reap its PID via ``os.waitpid``.

### Step 4 — OS-level pdeathsig opt-in (Issue 3, Linux only)

New ``pdeathsig_workers`` fixture sets ``PR_SET_PDEATHSIG = SIGKILL`` on every multiprocessing child via ``monkeypatch`` of ``BaseProcess._bootstrap``. Belt-and-suspenders for the case where pytest itself is SIGKILL'd before any in-process cleanup hook can run. No-op on non-Linux platforms.

### Issue 2 — flush stdout/stderr before os._exit

``os._exit`` doesn't flush block-buffered streams. Two-line patch adds ``sys.stdout.flush()`` + ``sys.stderr.flush()`` (inside try/except) before ``os._exit(0)``. Reproducer: a synthetic test that leaves a non-daemon thread alive at session end forced the os._exit branch and produced output ending at ``[100%]`` with no summary line (exit 0). With the fix, the same reproducer reports ``34 passed, 1 skipped in 1.29s`` correctly.

## Verification

- ``-m unit`` under JuniperCascor1: **3213 passed, 25 skipped, 0 failed**, summary line visible
- After-suite ``pgrep`` for ``JuniperCascor1.*forkserver`` shows zero NEW orphans from this run
- Synthetic non-daemon-thread reproducer now reports the summary line; pre-fix it didn't

## What's NOT in this PR

Step 5 — the standalone ``util/reap_pytest_orphans.bash`` script for cleaning up orphans *already* on the system from previous crashed sessions — ships separately as juniper-ml#196 (companion PR). Both PRs together close Issue 3 end-to-end.

## Test plan

- [x] Unit suite passes under JuniperCascor1
- [x] Zero new forkserver orphans after the run
- [x] Synthetic non-daemon-thread reproducer surfaces summary line post-fix
- [ ] Performance tests (``-m performance``) — would be nice to verify; not run here because they need significant runtime budget

### DIFF
--- a/src/tests/conftest.py
+++ b/src/tests/conftest.py
@@ -992,6 +992,22 @@ def pytest_unconfigure(config):
     alive = [t for t in threading.enumerate() if not t.daemon and t is not threading.main_thread()]
     if alive:
         import os
+        import sys
+
+        # Issue 2: ``os._exit`` skips ``atexit``, finalizers, AND the
+        # implicit stdout/stderr flush that normal Python exit performs.
+        # When pytest's stdout is redirected to a file (CI, ``> log``,
+        # tee, etc.) the streams are block-buffered, so the freshly-
+        # written ``X passed in Ys`` summary line and any post-summary
+        # diagnostic output sits in the buffer and is silently
+        # discarded the moment ``os._exit`` fires. Flush explicitly so
+        # the user sees the same final output whether or not the
+        # non-daemon-thread escape hatch was needed.
+        try:
+            sys.stdout.flush()
+            sys.stderr.flush()
+        except Exception:  # nosec B110 — never let cleanup raise here
+            pass
 
         os._exit(0)
 
--- a/src/tests/conftest.py
+++ b/src/tests/conftest.py
@@ -855,6 +855,53 @@ def _reap_multiprocessing_children(timeout: float = 2.0) -> None:
         pass
 
 
+@pytest.fixture
+def pdeathsig_workers(monkeypatch):
+    """Opt-in Linux-only defense: set ``PR_SET_PDEATHSIG = SIGKILL`` on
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #206, opened 2026-05-03
fix: pre-commit B014 + obs-wire-01 validation-loss gauge index (P-9, P-23)

pcalnon:
Two CI fixes blocking cascor main:

**P-9** — Pre-commit B014: `except (ChildProcessError, OSError)` is redundant since ChildProcessError inherits from OSError. From PR #205. Drop ChildProcessError.

**P-23** — Validation-loss/accuracy gauge stays at 0.0 when val history is shorter than train history (typical case — val runs every K epochs). The condition `last < len(val_loss_list)` evaluates false when the latest train epoch had no validation pass. Gauges are most-recent-observation; emit the last validation entry whenever any exists.

Surfaced together because P-9 unblocks pre-commit which unblocks unit-tests which surfaces P-23.


pcalnon (commented):
approved

PR #212, opened 2026-05-05
ci(lockfile): adopt constraint-satisfaction model + pin uv 0.11.8

pcalnon:
## Summary
- Replace the byte-for-byte Lockfile Freshness gate with the proven juniper-data constraint-satisfaction pattern (commit 1fca801, 2026-04-25)
- Pin \`uv==0.11.8\` in both \`ci.yml\` and \`lockfile-update.yml\` (was unpinned)
- Align Python version between regen (\`lockfile-update.yml\`: 3.12 → 3.14) and check (already 3.14)
- Regenerate \`requirements.lock\` to pick up sentry-sdk 2.59.0 (the trigger for this PR — main went red 2026-05-04 after PyPI shipped this patch)

## Why
The byte-for-byte diff fails when **any** transitive dependency releases a new patch on PyPI between the lockfile commit and the next CI run, even though no engineer changed anything in the repo. This false-positive failure mode has hit cascor (sentry 2.58→2.59, 2026-05-04), cascor-worker (cuda-pathfinder 1.5.1→1.5.4, 2026-05-03), canopy (data-client 0.4.0→0.4.1, 2026-05-03), and data (multiple times in April). juniper-data solved it in 1fca801 by switching to constraint-satisfaction. This PR propagates that fix.

The constraint-satisfaction model passes when the pinned versions still satisfy \`pyproject.toml\` and only fails when \`pyproject.toml\` drifts beyond what the lock can support (new dep added, min bumped past pin, entry removed). It tolerates transient PyPI churn.

See juniper-ml \`notes/LOCKFILE_FRESHNESS_REDESIGN_2026-05-04.md\` for the full ecosystem-wide analysis and rollout plan.

## Test plan
- [ ] CI green on this PR (in particular: \`Lockfile Freshness\` passes with constraint-mode + uv 0.11.8)
- [ ] Confirm post-merge \`main\` CI is green (the \`sentry-sdk\` failure mode resolved)
- [ ] Verify locally: \`uv pip compile pyproject.toml --extra ml --extra api --extra observability --extra juniper-data --index-strategy unsafe-best-match --no-emit-package torch --constraint requirements.lock -o /tmp/c.lock && diff <(grep '^[^[:space:]#]' requirements.lock | sort) <(grep '^[^[:space:]#]' /tmp/c.lock | sort)\` — empty diff = pass


pcalnon (commented):
approved

PR #215, opened 2026-05-05
feat(network): DELETE /v1/network/hidden-units/{idx} with cascade rebuild (Phase 6E CAN-015h-3) [retarget #201, stacked on #214]

pcalnon:
## Summary

**Retarget of #201, stacked on #214 (the h-2 retarget).** Same
incident as h-2: PR #201 merged into the stacked parent
\`feature/can-015h-2-add-hidden-unit-endpoint\` rather than \`main\`,
so commit \`0f67d114\` is **not an ancestor of \`origin/main\`** and
the \`DELETE /v1/network/hidden-units/{idx}\` endpoint has been
silently absent from \`main\` since 2026-05-03.

This is a clean cherry-pick of the original head commit \`9215e5c\`
onto the h-2 retarget branch (\`feature/can-015h-2-retarget\`).
The cherry-pick auto-merged \`src/api/lifecycle/manager.py\` with
no conflicts.

**Retarget to \`main\` after #214 lands.**

## Test plan

- [x] \`pytest tests/unit/api/test_remove_hidden_unit_manual.py\` —
      19 cases pass on the stacked branch.
- [x] Combined run with #214's tests:
      \`pytest test_add_hidden_unit_manual.py test_remove_hidden_unit_manual.py\`
      — 40 cases pass.
- [ ] CI green after retarget.

## Lineage

- Original PR: #201 (merged 2026-05-03 into the wrong base)
- Original head SHA: \`9215e5c\`
- Original merge SHA on stacked parent: \`0f67d114\` (orphaned w.r.t. \`main\`)
- Cherry-picked SHA on this branch: \`d5ebf5f\`

## Note on stacking

This PR's base is the h-2 retarget branch (#214) so the diff
shows only h-3 changes. After #214 merges to \`main\`, this PR
needs to be retargeted to \`main\` via \`gh pr edit --base main\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #214, opened 2026-05-05
feat(network): POST /v1/network/hidden-units with cascade rebuild (Phase 6E CAN-015h-2) [retarget #200]

pcalnon:
## Summary

**Retarget of #200 onto \`main\`.** PR #200 merged into the stacked
parent branch \`feature/can-015h-1-patch-weights-endpoint\` rather
than \`main\` (same incident class as g-2 #184 → #189 and g-3 #187 →
#190 earlier in the sprint), so commit \`f009063b\` is **not an
ancestor of \`origin/main\`** and the \`POST /v1/network/hidden-units\`
endpoint has been silently absent from \`main\` since 2026-05-03.

This brings the h-2 endpoint forward as a clean cherry-pick of the
original head commit \`69a4122\` onto current \`main\`. The
cherry-pick auto-merged \`src/api/lifecycle/manager.py\` against
post-h-1 main with no conflicts.

## Why now

Surfaced during a Phase 6E hardening pass: a round-trip
integration test (snapshot → restore → mutate → re-snapshot →
resume) couldn't run because the cascor side was missing the
mutation endpoint. \`grep '@router.' src/api/routes/network.py\`
on \`main\` showed only PATCH /weights — neither POST hidden-units
nor DELETE hidden-units/{idx} (h-3 #201, separate retarget PR
incoming).

User-visible impact while h-2/h-3 were missing: the canopy
**Network Editor** panel's \"Append unit\" and \"Delete unit\"
buttons returned 404 in service mode against any cascor build
from \`main\`. Demo mode masked the issue; the canopy unit tests
also masked it because they mocked the adapter.

## Test plan

- [x] \`pytest tests/unit/api/test_add_hidden_unit_manual.py\` —
      21 cases pass (FSM gate, validation, append, optimizer drop,
      tail-position, etc.) on the cherry-picked branch.
- [x] \`grep '@router.' src/api/routes/network.py\` shows
      \`@router.post(\"/hidden-units\")\` is now present.
- [ ] CI green.

## Lineage

- Original PR: #200 (merged 2026-05-03 into the wrong base)
- Original head SHA: \`69a4122\`
- Original merge SHA on stacked parent: \`f009063b\` (orphaned
  w.r.t. \`main\`)
- Cherry-picked SHA on this branch: \`5641cbd\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #219, opened 2026-05-05
fix(snapshots): coerce numpy scalars in get_training_params so /restore + /resume + /retrain return 200

pcalnon:
## Summary

Pre-existing bug surfaced during the Phase 6E hardening pass:
**every snapshot operation that returns the unified response
shape (\`/restore\`, \`/resume\`, \`/retrain\`) has been failing
in service mode with HTTP 400 since the response shape added
\`training_params\` (CAN-015d era).**

## Reproduce on \`main\` (any commit since CAN-015d)

\`\`\`
POST /v1/network → 200
POST /v1/snapshots → 200
POST /v1/snapshots/{id}/restore → 400
{
  \"status\": \"error\",
  \"error\": {
    \"code\": \"VALIDATION_ERROR\",
    \"message\": \"Invalid request parameters\",
    \"detail\": null
  }
}
\`\`\`

## Root cause

After \`load_snapshot\` reads the network's scalar attributes back
from HDF5, **h5py returns them as numpy scalars**
(\`numpy.int64\` / \`numpy.float64\`) rather than Python natives.
\`get_training_params()\` happily forwards those values into the
unified-response payload. Pydantic-core's JSON serializer then
rejects with:

\`\`\`
PydanticSerializationError: Unable to serialize unknown type: <class 'numpy.int64'>
\`\`\`

\`PydanticSerializationError\` is a \`ValueError\` subclass, so the
cascor app's \`value_error_handler\` catches it and returns
\`error_response(\"VALIDATION_ERROR\", \"Invalid request parameters\")\`
— without the \`detail=str(exc)\` argument. That's why the bug
has been silent: the response *looks* like a body-validation
failure, but pydantic body validation runs before the route
handler, so the user has no path to debug it.

## Why now

Surfaced while writing an end-to-end round-trip integration
test on cascor (snapshot → restore → mutate → re-snapshot →
resume). Existing snapshot tests exercise the lifecycle methods
directly, not the JSON-serialization path through the route
layer, so the regression slipped past them. **The canopy
\`Network Editor\` panel has been broken in service mode for the
same reason** — it depends on a successful restore to enter
Investigating state.

## Fix

- New \`_coerce_native_scalars(value)\` helper at module scope in
  \`lifecycle/manager.py\`. Walks dicts and lists/tuples, and for
  leaf values tries \`.item()\` — the duck-type interface numpy
  scalars and 0-d numpy arrays implement to return a Python
  native. Plain Python scalars don't have \`.item()\` and pass
  through unchanged.
- \`get_training_params\` wraps its return through the helper. No
  callers change shape — the values are already numerically
  equal, just typed correctly now.
- \`_build_unified_payload\` keeps its catch-all defensive except
  so a future regression here just logs rather than failing the
  operation entirely.

## Test plan

- [x] \`tests/unit/api/test_coerce_native_scalars.py\` — 20 cases
      covering numpy scalars (\`int8/16/32/64\`, \`float32/64\`,
      \`bool\`, 0-d arrays), containers (dict / nested dict /
      list / tuple, types preserved), passthrough for
      str/int/float/None/bool/empty, and the JSON round-trip the
      bug was actually about.
- [x] Manual verification post-fix:
      \`POST /v1/snapshots/{id}/restore → 200\`
      \`operation=restore, fsm_state=INVESTIGATING\`.
- [ ] CI green.

## Pairs with

- This is the third fix surfaced while landing the Phase 6E
  hardening pass. juniper-cascor #214 (h-2 retarget) and #215
  (h-3 retarget) brought the missing mutation endpoints onto
  main; this PR brings the snapshot ops back to working in
  service mode. The CAN-015h round-trip integration test is
  blocked on this PR.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -61,6 +61,43 @@ def _write_activation_function_name(network: Any, value: str) -> None:
     network._init_activation_function()
 
 
+def _coerce_native_scalars(value: Any) -> Any:
+    """Coerce numpy scalar types to Python natives, recursively.
+
+    pydantic-core's JSON serializer rejects ``numpy.int64`` /
+    ``numpy.float64`` with ``PydanticSerializationError: Unable to
+    serialize unknown type: <class 'numpy.int64'>``. After
+    ``load_snapshot`` the network's scalar attributes come back from
+    HDF5 as numpy scalars (h5py's default for attribute reads), so any
+    response payload that includes those values would fail
+    serialization with an opaque 400 (the cascor ``value_error_handler``
+    strips the detail).
+
+    Walks dicts and lists/tuples; pass-throughs anything that doesn't
+    expose ``.item()``. Used at the API surface (e.g.
+    ``get_training_params``) so the rest of the code can keep working
+    with numpy values internally.
+    """
+    if isinstance(value, dict):
+        return {k: _coerce_native_scalars(v) for k, v in value.items()}
+    if isinstance(value, (list, tuple)):
+        coerced = [_coerce_native_scalars(v) for v in value]
+        return type(value)(coerced) if isinstance(value, tuple) else coerced
+    item = getattr(value, "item", None)
+    if callable(item):
+        # numpy scalars and 0-d numpy arrays expose ``.item()``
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #222, opened 2026-05-05
fix(api): apply numpy-scalar coercion at the response envelope (CAN-015h follow-up to #219)

pcalnon:
## Summary

#219 patched a single call site (\`get_training_params\`), but the
underlying bug class — pydantic-core rejecting numpy scalars in
response payloads → 400 \`VALIDATION_ERROR\` with stripped detail —
affected more endpoints than that fix covered.

Hardening pass turned up at least:

| Endpoint | Pre-#219 | Post-#219 | Post-this-PR |
|---|---|---|---|
| \`POST /v1/snapshots/{id}/restore\` | 400 | 200 | 200 |
| \`POST /v1/snapshots/{id}/resume\` | 400 | 200 | 200 |
| \`POST /v1/snapshots/{id}/retrain\` | 400 | 200 | 200 |
| \`GET /v1/network\` (post-restore) | 400 | **400** | 200 |
| \`PATCH /v1/network/weights\` (post-restore) | 400 | **400** | 200 |

Both \`GET /v1/network\` and \`PATCH /v1/network/weights\` are
user-visible from the canopy Network Editor in service mode, so
this is in the same impact bucket as #219.

## Fix

Move the coercion up the stack so every route is covered.

- Promoted \`_coerce_native_scalars\` from \`api/lifecycle/manager.py\`
  (private, added by #219) to \`api/models/common.py\` as
  **\`coerce_native_scalars\`** (public).
- \`success_response(data)\` now runs \`data\` through the helper
  before returning the envelope. Every route's payload is
  JSON-clean regardless of whether the route author knew to coerce
  per-field.
- Kept a local \`_coerce_native_scalars\` alias in
  \`lifecycle/manager.py\` so existing internal callers don't have
  to be edited in this PR (the \`get_training_params\` wrapper
  added by #219 keeps working unchanged — both belt and braces).

## Why apply at the envelope rather than per-field

**Defence in depth.** Post-restore numpy scalars can leak into
payloads via paths that aren't obvious at the route layer (e.g.
nested inside \`hidden_units[*].bias\` from the topology builder).
Trying to enumerate every code path is fragile. The envelope is
the narrow waist where every successful response passes through,
and the helper is a no-op for already-Python-native data so the
cost is negligible.

## Test plan

- [x] \`test_success_response_coercion.py\` — 11 cases covering
      envelope shape preservation, numpy int/float at various
      nesting depths, passthrough for native data, \`error_response\`
      intentionally untouched, and a smoke test for the public
      import location.
- [x] Existing 20 cases in \`test_coerce_native_scalars.py\` still
      pass via the alias; total 31 cases green.
- [x] Pre-commit hooks pass.
- [x] Manual reproduction confirms \`GET /v1/network\` and
      \`PATCH /v1/network/weights\` flip 400 → 200 post-restore.
- [x] CI green.

## Pairs with

This is the fourth fix surfaced while landing the Phase 6E
hardening pass. Once this lands, the CAN-015h round-trip
integration test (which has been blocked on this serialization
issue) can finally land too.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -20,6 +20,7 @@ import torch
 
 from api.lifecycle.monitor import TrainingMonitor, TrainingState
 from api.lifecycle.state_machine import Command, TrainingPhase, TrainingStateMachine
+from api.models.common import coerce_native_scalars as _common_coerce_native_scalars
 from api.observability import TRAINING_SESSION_STATUS_CANCELLED, TRAINING_SESSION_STATUS_FAILURE, TRAINING_SESSION_STATUS_SUCCESS, dec_training_sessions, inc_training_session_completed, inc_training_sessions, observe_training_step_duration, record_training_epoch, set_hidden_units, set_training_accuracy, set_training_loss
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #223, opened 2026-05-06
fix(ci,tests): integration tests actually run + register critical mark

pcalnon:
## Summary
- **Pass \`--integration\` in both Quick + Full Integration job invocations** so the conftest skip-gate at \`src/tests/conftest.py:224\` lets the tests actually execute.
- **Register \`critical\` marker** in \`[tool.pytest.ini_options].markers\` to silence \`PytestUnknownMarkWarning\` from \`src/tests/integration/api/test_ws_control_set_params.py:72\`.

## Why
Both Integration Tests jobs in the latest main CI run reported \`60 skipped, 21 deselected\` and finished in **~1.5s** — i.e. **zero tests executed**. \`conftest.py\` adds a skip marker to every integration-marked test unless \`--integration\` is passed on the command line. The CI invocation never passed it, so both jobs silently became no-ops since the gate was introduced.

The unit-test job runs against \`src/tests/unit\` with \`-m \"unit and not slow\"\`, so no spillover risk from passing \`--integration\` only in the integration jobs.

## Test plan
- [ ] CI green on this PR — and \`Quick Integration Tests\` + \`Full Integration Tests\` jobs report >0 tests run (will reveal any pre-existing integration test failures that were hidden by the silent skip)
- [ ] \`PytestUnknownMarkWarning\` for \`critical\` no longer appears in test output

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,7 +285,11 @@ jobs:
           echo "║       Juniper Cascor - Quick Integration Tests             ║"
           echo "╚════════════════════════════════════════════════════════════╝"
 
+          # --integration opts in to integration-marked tests; without it
+          # src/tests/conftest.py adds a skip marker to every integration
+          # test, which silently turned this entire job into a no-op.
           python -m pytest \
+            --integration \
             -m "integration and not slow" \
             src/tests/integration \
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #225, opened 2026-05-06
test(api): CAN-015h round-trip integration — snapshot → restore → mutate → re-snapshot → resume → train

pcalnon:
## Summary

Adds the integration test called for in the CAN-015h design plan's
**Med risk row**:

> *Optimizer rebuild after add/remove unit corrupts subsequent
> training.* → Round-trip integration test: snapshot → restore →
> add unit → re-snapshot → resume training → verify training
> converges.

This is the deliverable for the Phase 6E CAN-015h hardening pass.
The four pre-existing bugs surfaced while writing this test all
landed first — see "Pairs with" below — without which the test
would have been chasing them as red herrings.

## Coverage

\`test_full_roundtrip_add_remove_patch\` walks every CAN-015h
surface through the live FastAPI TestClient (no external server):

| # | Step | Asserts |
|---|---|---|
| 1 | POST /v1/network + capture topology | network created, pre-snapshot topology readable |
| 2 | POST /v1/snapshots | snapshot id + path returned |
| 3 | POST /v1/snapshots/{id}/restore | unified body + FSM = Investigating (verified independently via /v1/training/status) |
| 4 | PATCH /v1/network/weights (output bias) | h-1 path, 200 |
| 5 | POST /v1/network/hidden-units | h-2 tail-append, num_hidden + 1 |
| 6 | DELETE /v1/network/hidden-units/{idx} | h-3 cascade-rebuild, num_hidden restored |
| 7 | POST /v1/snapshots (re-snapshot) | post-mutation network is serializable; id ≠ initial |
| 8 | POST /v1/snapshots/{id}/resume | unified body + FSM = RESUME_READY |
| 9 | POST /v1/training/start | optimizer rebuild fires here |
| 10 | GET /v1/metrics | post-resume train_loss is finite, non-NaN |

\`test_mutations_rejected_outside_investigating\` is the targeted
FSM-gate sanity: all three mutation routes must return 409 from
any state that isn't Investigating.

## Why integration vs. unit

Per-route unit tests (\`test_patch_weights_manual\`,
\`test_add_hidden_unit_manual\`, \`test_remove_hidden_unit_manual\`)
already cover behaviour-in-isolation. What's *not* covered there
is the **interaction**:

- Does the optimizer rebuild correctly when training resumes after
  add/remove? (Optimizer state is dropped by h-1/h-2/h-3.)
- Does the cascade-rebuild surgery in \`remove_hidden_unit_manual\`
  produce a network that \`resume_from_snapshot\` can re-load and
  \`train\` can drive forward?
- Does the FSM gate the restored-and-mutated state correctly?

This is the test that catches \"the unit tests passed but the
end-to-end flow is broken\" regressions.

## Test plan

- [x] Both cases pass locally:
      \`pytest tests/integration/api/test_can_015h_roundtrip.py --integration\`
      → 2 passed in 13.16s.
- [x] Pre-commit hooks pass.
- [ ] CI green.

## Notes

- The test deliberately does **not** train before the initial
  snapshot. The optimizer-rebuild concern fires at the resume-train
  step regardless of whether the snapshotted network had a prior
  training pass; skipping the initial train keeps the test under a
  reasonable wall-clock budget (~13s).
- Cascor schema for PATCH is \`target ∈ {output, hidden_unit}\` +
  \`field ∈ {weights, bias}\` per
  \`api/models/network.py::PatchWeightsRequest\` — the test sends
  the cascor schema directly.
- Snapshot ids are at second precision so the test sleeps 1.1s
  between the two snapshot ops to avoid id collision. Tracked
  separately as a small lifecycle-improvement opportunity.

## Pairs with

This is the closing PR of the Phase 6E hardening pass. The four
bugs surfaced and fixed during the pass:

1. juniper-cascor #214 — h-2 retarget (POST /v1/network/hidden-units missing from main)
2. juniper-cascor #215 — h-3 retarget (DELETE /v1/network/hidden-units/{idx} missing from main)
3. juniper-canopy #239 — Network Editor PATCH target schema mismatch
4. juniper-cascor #219 + #222 — pydantic-core rejecting numpy scalars (narrow then envelope-level fix)

Plan: \`juniper-ml/notes/PHASE_6E_DEFERRED_CAN-015GH_DESIGN.md\`
§\"Risks — CAN-015h\".

### DIFF
--- a/src/tests/integration/api/test_can_015h_roundtrip.py
+++ b/src/tests/integration/api/test_can_015h_roundtrip.py
@@ -1,0 +1,358 @@
+"""CAN-015h round-trip integration test.
+
+Mitigates the design plan's "Med" risk row:
+
+    Optimizer rebuild after add/remove unit corrupts subsequent
+    training. → Round-trip integration test: snapshot → restore →
+    add unit → re-snapshot → resume training → verify training
+    converges.
+
+Scope: walk the full snapshot/restore/mutate/re-snapshot/resume
+pipeline through the live FastAPI surface (``TestClient``, no
+external server). Each step asserts both the HTTP contract (status
+code + response shape) and the FSM transition the operation is
+supposed to drive, so a regression in either layer trips the test.
+
+The test is deliberately *integration*-flavored — unit-level
+behaviour for each individual mutation is already covered by
+``test_patch_weights_manual``, ``test_add_hidden_unit_manual``, and
+``test_remove_hidden_unit_manual``. What's *not* covered there is
+the interaction:
+
+- Does the optimizer rebuild correctly when training resumes after
+  add/remove? (Optimizer state is dropped by h-1/h-2/h-3.)
+- Does the cascade-rebuild surgery in ``remove_hidden_unit_manual``
+  produce a network that ``resume_from_snapshot`` can re-load and
+  ``train`` can drive forward?
+- Does the FSM gate the restored-and-mutated state correctly so the
+  user can't accidentally trigger ``start_training`` while still in
+  ``Investigating``?
+
+This is the test that catches "the unit tests passed but the
+end-to-end flow is broken" regressions.
+"""
+
+import threading
+import time
+from pathlib import Path
+
+import pytest
+from fastapi.testclient import TestClient
+
+from api.app import create_app
+from api.settings import Settings
+
+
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #228, opened 2026-05-06
fix(utils): load_dataset uses torch.load(weights_only=True), not yaml (BUG-CC-12)

pcalnon:
## Summary

\`utils.load_dataset\` was supposed to invert \`utils.save_dataset\`
(which writes via \`torch.save\`), but the active path was
\`yaml.safe_load(file_path.read())\` — a half-finished format swap
that never round-tripped a real \`torch.save\` payload. The two
functions had not been an inverse pair since the swap landed.

Surfaced by the 2026-05-05 roadmap audit
(\`notes/ROADMAP_AUDIT_2026-05-05.md\` §4.1, BUG-CC-12).

## Root cause

The existing tests passed because they fed YAML strings directly
into \`load_dataset\` and never round-tripped through
\`save_dataset\`. A type-hint mismatch (\`file_path: str\` vs.
\`.read()\` call) meant the function could only be called with a
file-like object — not the string path the signature advertised.

## Fix

- \`load_dataset\` now uses
  \`torch.load(file_path, map_location=\"cpu\", weights_only=True)\`.
- \`weights_only=True\` is the torch ≥ 2.6 default and required from
  torch 2.8; it restricts the unpickler to safe primitive types.
- \`file_path\` type hint widened to \`Union[str, os.PathLike[str]]\`.
- \`yaml\` import removed (this was its only caller).

## Test plan

Replaced two YAML-string tests (which encoded the broken behaviour)
with three round-trip cases:

- [x] \`test_load_dataset_round_trip\` — save_dataset → load_dataset
      returns equal tensors via \`torch.testing.assert_close\`.
- [x] \`test_load_dataset_returns_tensors\` — result is real
      \`torch.Tensor\` instances (the YAML path lost type fidelity).
- [x] \`test_load_dataset_accepts_pathlib\` — \`pathlib.Path\` works.
- [x] 49 cases pass in \`test_utils_coverage.py\` post-fix.
- [x] Pre-commit hooks pass.

### DIFF
--- a/src/tests/unit/test_utils_coverage.py
+++ b/src/tests/unit/test_utils_coverage.py
@@ -2,7 +2,6 @@
 """
 Tests for utils/utils.py to increase code coverage.
 """
-import io
 import os
 import sys
 import tempfile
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #229, opened 2026-05-06
ci(workflow): force gitleaks-action onto Node.js 24

pcalnon:
## Summary
Add `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"` to the gitleaks-action env block.

## Why
GitHub forces Node.js 20 actions onto Node.js 24 on **2026-06-02**. gitleaks-action v2.3.9 has no Node.js 24 release (upstream PR #215 in flight). The documented escape hatch eliminates the deprecation warning now and protects against the upcoming cutover. Drop the override once an upstream node24-pinned release ships.

## Test plan
- [ ] CI green
- [ ] `Node.js 20 actions are deprecated` warning gone from Security Scans output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #230, opened 2026-05-06
fix(ci): skip gitleaks on repository_dispatch events

pcalnon:
## Summary
Add \`if: github.event_name != 'repository_dispatch'\` to cascor's gitleaks-action step. Mirrors the guard already present in juniper-data and juniper-canopy.

## Why
After PR #229 (gitleaks-action Node.js 24 forcing) merged, cascor main went red on the Security Scans job:

\`\`\`
##[error]ERROR: The [repository_dispatch] event is not yet supported
\`\`\`

cascor subscribes to \`data-client-updated\` and \`cascor-worker-updated\` dispatch events. The ecosystem-merge cascade from this evening's six gitleaks PRs fired those events into cascor's CI, and gitleaks-action v2.3.9 explicitly rejects \`repository_dispatch\`. The push/PR triggers still cover human-driven entry points; the dispatch trigger is for ecosystem-fanout test runs and never introduces new commits.

The pre-existing bug went undetected because \`repository_dispatch\` rarely fires; this evening's batch of merges happened to fan out fast enough to trigger it.

## Test plan
- [ ] CI green on this PR (push event — gitleaks runs as before)
- [ ] Post-merge: any future \`repository_dispatch\`-triggered cascor CI run skips Gitleaks step cleanly without taking down Security Scans

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,6 +482,16 @@ jobs:
           fetch-depth: 0
 
       - name: Run Gitleaks (Secrets Detection)
+        # gitleaks-action v2.3.9 explicitly rejects ``repository_dispatch``
+        # ("ERROR: The [repository_dispatch] event is not yet supported"),
+        # which kills the entire Security Scans job whenever the workflow
+        # is triggered by upstream-published events from juniper-data /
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #233, opened 2026-05-06
refactor(observability): migrate to juniper-observability register_or_reuse (Phase 2c)

pcalnon:
## Summary

Phase 2c — final per-consumer migration of the plan in [`juniper-ml/notes/observability/REGISTER_OR_REUSE_HELPER_DESIGN_2026-05-05.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/observability/REGISTER_OR_REUSE_HELPER_DESIGN_2026-05-05.md).

## What changes

| File | Change |
|---|---|
| ``src/api/observability.py`` | Drop the local ``_register_or_reuse`` definition (~32 lines) and replace with ``from juniper_observability import register_or_reuse as _register_or_reuse``. The 22+ existing call sites in ``_ensure_training_metrics`` / ``_ensure_ws_metrics`` keep their exact same syntax via the alias. |
| ``src/api/websocket/control_stream.py`` | ``_get_command_counter`` swaps inline try/except for the imported helper. Net −15 lines. |
| ``pyproject.toml`` | ``juniper-observability>=0.1.1`` → ``>=0.2.0`` |

## Semantic note: drop-and-recreate vs adopt-existing

The original local ``_register_or_reuse`` used **drop-and-recreate** on a duplicate (unregister existing → retry construction → latest args win at the cost of discarded samples). The canonical ``juniper_observability.register_or_reuse`` uses **adopt-existing** (samples preserved, latest args silently ignored).

For all 22+ call sites in this module the args are identical every time (the lazy-init singleton dict pattern), so the two semantics produce indistinguishable output for production *and* for tests that start with a clean REGISTRY. Adopt-existing is strictly more correct for any scenario where a fixture clears the module-globals without scrubbing REGISTRY itself.

If a future call site needs drop-and-recreate, ``juniper_observability.register_fresh`` is available with the old semantics.

## Verification

Full ``-m unit`` cascor suite under JuniperCascor1: **3346 passed, 10 skipped, 0 failed**.

## Companion PRs

- juniper-data #95 (Phase 2a) — opened
- juniper-canopy #245 (Phase 2b) — opened
- juniper-cascor-client deliberately excluded from Phase 2 (would add pydantic + starlette transitive deps to a leaf library; PR #37 inline guard stays)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
     # configure_sentry, get_prometheus_app, set_build_info, and the
     # R1.1/R1.2/R1.3 contract constants). Pinned at the first stable
     # release; consumers should never resolve back to the alpha line.
-    "juniper-observability>=0.1.1",
+    "juniper-observability>=0.2.0",
     # METRICS-MON R2.2.2 / seed-05: WebSocket frame schemas (envelope
     # Pydantic models for /ws/training and /ws/control + WorkerMessageType
     # StrEnum + BinaryFrame numpy codec for /ws/v1/workers). The cascor
--- a/src/api/observability.py
+++ b/src/api/observability.py
@@ -171,38 +171,27 @@ _TRAINING_STEP_DURATION_BUCKETS: tuple = (
 )
 
 
-def _register_or_reuse(cls, name: str, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #236, opened 2026-05-07
fix(tests): add pytest-benchmark to test extra so perf tests resolve `benchmark` fixture

pcalnon:
## Summary

- Adds `pytest-benchmark>=4.0.0` to the `[project.optional-dependencies] test` extra in `pyproject.toml`. Performance tests under `src/tests/performance/` use the `benchmark` fixture provided by pytest-benchmark, but the dep was missing from the test extra. Running `pytest --slow --run-performance --integration -vvv --cov .` therefore errored out 82 perf tests with `fixture 'benchmark' not found` while the standard `pytest -vvv --cov .` invocation still passed.
- Adds `src/tests/unit/test_pytest_benchmark_available.py` — a unit-scope regression guard (3 assertions: import, `pytest11` entry-point registered, fixture resolvable) that fails fast with an actionable message pointing at the test extra in `pyproject.toml` if the dep ever silently drops out again, instead of cascading dozens of confusing fixture errors at `--run-performance` time.

## Test plan

- [x] Full failing command reproduces clean: `pytest --slow --run-performance --integration -vvv --cov .` → **3683 passed, 0 errors, 1 skipped, 3 subtests passed in 332.42s** (was 3598 passed, **82 errors**, 1 skipped).
- [x] The 3 new regression tests pass against the fixed env (`JuniperCascor1`, Python 3.13, pytest 9.0.3, pytest-benchmark 5.2.3).
- [ ] Standard `pytest -vvv --cov .` still passes in CI.
- [ ] CI smoke (build + extras metadata) confirms the new dep resolves on supported Python versions.


pcalnon (commented):
approved

pcalnon (commented):
approved

pcalnon (commented):
approved

PR #237, opened 2026-05-07
chore(deps): refresh requirements.lock for juniper-observability 0.2.0 bump

pcalnon:
## Summary

- `pyproject.toml` requires `juniper-observability>=0.2.0` (since the `register_or_reuse` migration on 2026-05-06), but `requirements.lock` still pinned `juniper-observability==0.1.1`. The constraint-mode `Lockfile Freshness` gate (introduced 2026-05-04, see [memory note on the constraint-satisfaction redesign](https://github.com/pcalnon/juniper-cascor/blob/main/AGENTS.md)) correctly fails this as **unsatisfiable**, blocking every PR opened against `main` since the bump.
- Regenerates `requirements.lock` on Python 3.14 (matching `PYTHON_TEST_VERSION` and `lockfile-update.yml`) with `uv 0.11.8` and the canonical compile-to-`/tmp` + `mv` pattern from `lockfile-update.yml` (avoids the [uv self-pin trap](https://github.com/pcalnon/juniper-ml) where uv reads the existing `-o` target as soft constraints).

## Diff

Three line changes total:

```diff
- juniper-observability==0.1.1
+ juniper-observability==0.2.0
- pydantic==2.13.3
+ pydantic==2.13.4
- pydantic-core==2.46.3
+ pydantic-core==2.46.4
```

The pydantic patches are transitive PyPI patch releases that shipped between the last regen and now — expected and benign under constraint-mode.

## Recurring-issue notes

This is the **third** distinct lockfile-related break since 2026-05-03. Each had a different mechanism:

| Date       | Mechanism                                                | Fix                                                 |
| ---------- | -------------------------------------------------------- | --------------------------------------------------- |
| 2026-05-03 | Byte-for-byte gate flagged transitive PyPI patches       | cascor#212 — switched to constraint-satisfaction    |
| 2026-05-04 | Local regen self-pinned via `-o requirements.lock`       | Memory note + `lockfile-update.yml` uses `/tmp`+mv  |
| 2026-05-06 | `pyproject.toml` min bumped without `requirements.lock` regen | **This PR**                                    |

The current break is **not** a regression of the gate itself — the gate worked. The miss was that the `juniper-observability>=0.2.0` PR (cascor migration to `register_or_reuse`) updated `pyproject.toml` without regenerating the lockfile, and the gate caught it on the next PR. Possible follow-ups:

- Add a one-line `Refresh requirements.lock if you bumped a dep min` reminder to `AGENTS.md` near the lockfile section, since this is the third footgun mode.
- (Larger) Consider an opt-in pre-commit hook that runs the constraint-mode check against `requirements.lock` so the local commit fails before CI does.

## Test plan

- [x] Local simulation of the CI gate (`uv pip compile … --constraint requirements.lock` + sorted-pins diff) passes against the refreshed lockfile.
- [ ] CI `Lockfile Freshness` job goes green on this PR.
- [ ] PR #236 (pytest-benchmark fix) re-runs green after this lands and is merged in.


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #238, opened 2026-05-09
fix(ws): broadcast full topology on cascade_add (not count-only stub)

pcalnon:
## Summary

The `topology` WS frame emitted after cascade growth was a metadata stub (`hidden_units` as int count). Canopy's `_transform_topology` expects a list of hidden-unit dicts; `isinstance(int, list) is False` silently collapsed the result to inputs+outputs only and zero hidden nodes, so the Network Topology view in canopy never updated as the cascade grew.

Replace the stub with `manager.get_topology()` — the same serializer the REST `/v1/network/topology` endpoint uses — so the WS frame carries `hidden_units: [{id, weights, bias, activation}, ...]` plus `output_weights` / `output_bias`. Existing `_topology_lock` is held by `get_topology()`, so threadsafety is preserved.

## Symptom (pre-fix)

Network Topology tab header showed `Hidden Units: 0` and `Total Connections: 4` (= 2 inputs × 2 outputs) even after the cascade had grown. The sidebar `Network Information` panel correctly showed `Hidden Units: 1` because it reads `len(network.hidden_units)` from `/api/status`, which is unaffected by this WS path.

Companion canopy PR adds defense-in-depth: pcalnon/juniper-canopy#fix/topology-ws-broadcast-completeness.

## Test plan

- [x] New `test_cascade_add_topology_broadcast_carries_full_serialized_network` in `src/tests/unit/api/test_monitoring_hooks.py` — exercises the monitored `grow_network` wrapper end-to-end and asserts the broadcast `data.hidden_units` is a list, each unit carries `weights`/`bias`/`activation`, and `output_weights` / `output_bias` are present so canopy's transform can build hidden→output edges.
- [x] All 1259 existing `tests/unit/api/` tests pass (including the existing `test_ws_callbacks_broadcast_on_cascade_add` for the separate `cascade_add`-typed broadcast).
- [ ] Live observation in canopy after merge — Network Topology tab should reflect new hidden nodes within one fast-update tick of cascade growth.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -1594,13 +1594,19 @@ class TrainingLifecycleManager:
                         correlation=actual_correlation,
                     )
                 if manager_ref._ws_manager is not None:
-                    topology_data = {
-                        "hidden_units": new_hidden,
-                        "input_size": getattr(manager_ref.network, "input_size", 0),
-                        "output_size": getattr(manager_ref.network, "output_size", 0),
-                        "event": "cascade_add",
-                    }
-                    manager_ref._ws_manager.broadcast_from_thread(create_topology_message(topology_data))
+                    # The cascade_add ``topology`` broadcast must carry the
+                    # full serialized network — hidden_units as a list of
+                    # per-unit dicts (weights/bias/activation) plus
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #239, opened 2026-05-10
fix(topology): handle ActivationWithDerivative in get_topology serializer

pcalnon:
## Summary

`manager.get_topology()` extracted activation names via `unit.get("activation_fn", torch.sigmoid).__name__`. That works for plain torch builtins (manual `add_hidden_unit` path) but breaks on the production path: `cascade_correlation.py` wraps every activation in an `ActivationWithDerivative` instance for multiprocessing picklability, and that wrapper exposes its name via `_activation_name`, not `__name__`.

## Symptom (pre-fix)

The `AttributeError` was swallowed by the surrounding `except Exception` block, so it produced no traceback anywhere visible — the route just returned `500 {"detail":"Failed to extract topology"}`. As a result:

1. `GET /v1/network/topology` returned 500 silently.
2. The cascade_add WS broadcast (#238 fix) skipped sending the topology frame because `get_topology()` returned None — `if full_topology is not None` short-circuit.
3. Canopy's Network Topology view stayed stuck on the initial 0-hidden snapshot for the rest of the session — the **same user-visible symptom** that motivated the original count-only WS stub investigation.

This is the third (innermost) layer of the topology-display bug. #238 fixed the WS broadcast shape; canopy PR #253 added defensive gates; this PR fixes the serializer that both of those rely on.

## Verification

Live on the dashboard after this fix:

| Source | Pre-fix | Post-fix |
|---|---|---|
| Sidebar `Network Information` → `Hidden Units` | 1 (correct, via `/v1/network`) | 1 |
| Topology tab `Hidden Units` | **0** (incorrect) | **1** (correct, via `/v1/network/topology`) |
| Topology tab `Total Connections` | **4** (= 2×2 inputs→outputs only) | **8** (= 2 input→hidden + 4 input→output + 2 hidden→output) |
| Plot rendering | only input + output nodes visible | hidden node + cascade edges visible |

## Test plan

- [x] `test_get_topology_with_hidden_unit_using_activation_wrapper` — injects a unit shaped exactly like a candidate-trained one (weights/bias as Tensors, `activation_fn=ActivationWithDerivative(torch.tanh)`) and asserts `get_topology()` returns a list-shaped payload with `activation == "tanh"`.
- [x] `test_activation_name_helper_handles_known_shapes` covers wrapper, builtin (`torch.sigmoid`/`torch.tanh`), `nn.Module` (`torch.nn.Tanh`/`ReLU`), `None`, and a bare class.
- [x] All 1261 existing `tests/unit/api/` tests pass (1259 prior + 2 new).
- [x] Live verified end-to-end via canopy dashboard with a freshly-trained network — topology tab now reflects added hidden units.


pcalnon (commented):
approved

PR #241, opened 2026-05-10
feat(params): candidate-pool PATCH schema + invariants (Issue #1, PR-4a)

pcalnon:
## Summary

PR-4a of the [FRONTEND_ISSUES_PLAN_2026-05-09 §1.5](https://github.com/pcalnon/juniper-canopy/blob/main/notes/FRONTEND_ISSUES_PLAN_2026-05-09.md) series — closes the canopy-adapter silent-drop for the five candidate-pool selection knobs by giving cascor a real schema, storage, and post-merge invariant validation for them. PR-5 (canopy adapter map extension) can now ship without 404 storms.

> **Scope is intentionally schema-only.** Spec §1.5 C2 implicitly assumed the candidate-pool selection concepts (multi/top/random) already existed in `cascade_correlation.py` — they don't. Splitting the work as PR-4a (this — schema + validation) and PR-4b (selection-logic rewrite in `_select_best_candidates`/`grow_network`) was confirmed in scoping. The five new fields round-trip through PATCH/GET and persist on the network today, but the live training loop still uses the legacy single-top-correlation path. Defaults preserve current behavior.

## Five new params

| Field | Type | Default | Purpose |
|------|------|---------|---------|
| `multi_candidate` | bool | False | Promote multiple candidates per growth iteration (PR-4b) |
| `candidate_selection` | Literal["top","random","mixed"] | "top" | Selection strategy (PR-4b) |
| `selected_candidates` | int >= 1 | 1 | S — total promoted per iteration |
| `top_candidates` | int >= 0 | 1 | T — top-correlation slice |
| `random_candidates` | int >= 0 | 0 | R — random slice |

## Wire-up

- `src/api/models/training.py`: extends `TrainingParamUpdateRequest` and `TrainingParams` with the five fields, mirroring existing pydantic constraint style.
- `src/cascade_correlation/cascade_correlation.py`: stores the five attributes on `CascadeCorrelationNetwork.__init__` with defaults that preserve current behavior.
- `src/api/lifecycle/manager.py`:
  - new module-level `InvalidCandidatePoolError(ValueError)` — subclasses `ValueError` so existing handlers keep working
  - new pure helper `_validate_candidate_pool_triple(s, t, r, p)` returning `None` or a violation string
  - new `TrainingLifecycleManager._validate_candidate_pool_post_merge(params)` that folds the PATCH delta over the current network values before validating
  - `_apply_params_unlocked` calls the validator before any setattr, so a violating PATCH never partially applies
  - `get_training_params` now surfaces the five fields
- `src/api/routes/training.py`: PATCH `/v1/training/params` translates `InvalidCandidatePoolError` to **HTTP 422** with the violation string in `detail` (distinct from 404 for missing network).

## Atomicity (Resolution log Q2 — required, not optional)

The validator runs against the **post-merge** triple, not the per-key delta. This means a multi-key PATCH that's only valid as a unit is accepted in one shot:

```http
PATCH /v1/training/params
{ "selected_candidates": 6, "top_candidates": 4, "random_candidates": 2 }
```

…starting from `{S=1, T=1, R=0}` succeeds, even though per-key (S=6 with inherited T=1, R=0) would fail `T+R == S`. A test pins this explicitly.

`candidate_pool_size` in the same PATCH validates against the new pool, not the old — also pinned by a test.

## Truth table

Per spec §1.5 C2.1:

1. `1 ≤ S ≤ P`
2. `T ≥ 0` and `R ≥ 0`
3. `T ≤ S` and `R ≤ S`
4. Degenerate: if `R == 0` then `T == S`; if `T == 0` then `R == S`
5. Both nonzero: `T + R == S`
6. `T == 0 and R == 0` is illegal when `S > 0`

## Test plan

- [x] `pytest src/tests/unit/api/test_param_validation_helper.py` — 17/17 (truth table)
- [x] `pytest src/tests/integration/api/test_candidate_pool_invariants.py --integration` — 18/18 (valid + invalid + atomicity + pool-size + per-field round-trip + Literal rejection)
- [x] `pytest src/tests/unit/api/test_lifecycle_*.py --integration` — 209/209 (no lifecycle regression)
- [x] `pytest src/tests/integration/api/ --integration` — 74/74 (no broader regression)

## Follow-ups

- **PR-4b**: rewire `_select_best_candidates` + `grow_network` to actually consume `multi_candidate` / `candidate_selection` / `(S, T, R)`. Adds end-to-end behavior tests showing the promotion logic respects the triple.
- **PR-5** (canopy): extend `_CANOPY_TO_CASCOR_PARAM_MAP` with the five new keys + the C2.2 clientside validator.
- **PR-4 spec §1.6 — additional canopy-side adapter mock-drop test** ships with PR-5.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -71,6 +71,43 @@ def _write_activation_function_name(network: Any, value: str) -> None:
 _coerce_native_scalars = _common_coerce_native_scalars
 
 
+class InvalidCandidatePoolError(ValueError):
+    """Raised when a candidate-pool PATCH violates the §1.5 C2.1 invariant triple.
+
+    A subclass of ``ValueError`` so existing ``except ValueError`` handlers in
+    ``update_params`` and the PATCH route still see it; the PATCH route promotes
+    instances of this specific subclass to HTTP 422 (the message is the
+    human-readable violation string) while bare ``ValueError`` remains 404.
+    See ``FRONTEND_ISSUES_PLAN_2026-05-09.md §1.5 C2.1`` for the full truth table.
+    """
+
+
+def _validate_candidate_pool_triple(s: int, t: int, r: int, p: int) -> Optional[str]:
+    """Return ``None`` if (S, T, R, P) satisfy the §1.5 C2.1 invariant or a
+    human-readable violation string otherwise.
+
+    Bound to be called against the *post-merge* triple — never against the
+    per-key delta — because a multi-key PATCH that's only valid as a unit
+    (e.g. ``{S: 6, T: 4, R: 2}`` from a prior ``(S=2, T=2, R=0)``) must be
+    accepted in one shot.
+    """
+    if not (1 <= s <= p):
+        return f"selected_candidates {s} not in [1, candidate_pool_size={p}]"
+    if t < 0 or r < 0:
+        return f"top_candidates and random_candidates must be >= 0 (got T={t}, R={r})"
+    if t > s or r > s:
+        return f"each component must be <= selected_candidates (S={s}, T={t}, R={r})"
+    if t == 0 and r == 0:
+        return "top_candidates and random_candidates cannot both be 0"
+    if t == 0 and r != s:
+        return f"with top_candidates=0, random_candidates must equal S={s} (got R={r})"
+    if r == 0 and t != s:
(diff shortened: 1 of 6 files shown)


pcalnon (commented):
approved

PR #244, opened 2026-05-10
fix(lifecycle): make pause_training and stop_training actually interrupt training

pcalnon:
## Summary

- **Defect fix.** Discovered during the P2-1a pause-boundary audit for the Phase 2 live-dataset-swap work (canopy `notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md` §3.4).
- Pre-fix, `pause_training` and `stop_training` updated the FSM and broadcast new state via WebSocket, but **training continued to natural completion regardless** — the events were never observed inside `cascade_correlation.fit()` or any inner training loop.
- Fix wires signal checks into the existing `_output_training_callback` (every 25 output epochs) and `_grow_iteration_callback` (cascade-iteration boundary) via a new `_check_for_interrupt()` instance method that raises `TrainingInterrupted` on stop and blocks on pause. `monitored_fit` catches `TrainingInterrupted` as a clean cancellation (same FSM/state/gauge transitions as the post-fit stop-event path).

## Why this slipped through

Existing tests (`test_lifecycle_event_invariants.py`, `test_lifecycle_manager.py::test_stop_training`) verified the *event states* and the *return dict shape* but never asserted that **training actually stops**. The new test file pins the missing behavior contract directly — see `test_monitored_fit_treats_interrupt_as_clean_cancellation` and the four `test_output_callback_*` tests.

## What's in this PR

**Production code** (`src/api/lifecycle/manager.py`):
- New `TrainingInterrupted(Exception)` sentinel near `InvalidCandidatePoolError`.
- New `_check_for_interrupt()` instance method (the missing pause/stop hook).
- Two existing callbacks now call `manager_ref._check_for_interrupt()`.
- `monitored_fit` gains a `except TrainingInterrupted` clause that catches cleanly (does NOT re-raise, increments `TRAINING_SESSION_STATUS_CANCELLED`).

**Refactor** (defensible on its own):
- `_output_epoch_callback` and `_grow_iteration_callback` are now bound to the network at install time, not on first fit. Removes a subtle race window.
- `_check_for_interrupt` is an instance method (not a closure) so it's reachable from `_install_grow_network_hook` (separate method scope).

**Tests** (`src/tests/integration/api/test_pause_stop_actually_interrupts.py`, 8 cases):
1. `test_output_callback_raises_on_stop_signal`
2. `test_output_callback_returns_normally_when_not_paused_or_stopped`
3. `test_output_callback_blocks_on_pause_then_resumes`
4. `test_output_callback_raises_on_stop_during_pause` (validates the 0.5s wait timeout — pre-fix this was impossible)
5. `test_grow_callback_raises_on_stop_signal`
6. `test_grow_callback_returns_normally_when_not_stopped`
7. `test_monitored_fit_treats_interrupt_as_clean_cancellation`
8. `test_monitored_fit_treats_other_exceptions_as_failure` (pins the boundary so a future refactor doesn't widen the cancellation catch)

## Scope notes

- **Pause inside multiprocessing candidate-training is intentionally out of scope.** The cascade-iteration boundary is the natural pause point for cascade growth; tighter granularity would require multiprocessing-aware signal threading.
- **No behavior change for completed training runs** — the post-fit stop-event check at `manager.py:1500` is preserved as a fallback (now redundant in the typical case but harmless).
- The full `src/tests/unit/api/ + src/tests/integration/api/` sweep passes after the change.

## Test plan

- [x] `pytest tests/integration/api/test_pause_stop_actually_interrupts.py --integration` — 8/8 pass
- [x] `pytest tests/unit/api/test_lifecycle_manager.py tests/integration/api/test_lifecycle_event_invariants.py tests/unit/test_cascade_correlation_callback_hooks.py --integration` — 130/130 existing tests pass (no regression)
- [x] `pytest tests/unit/api/ tests/integration/api/ --integration` — exit 0
- [x] `pre-commit run --files src/api/lifecycle/manager.py src/tests/integration/api/test_pause_stop_actually_interrupts.py` — all hooks pass (flake8, bandit, mypy, black, isort)
- [ ] Manual smoke test: start training in canopy, click Pause, observe training metrics stop emitting; click Stop, observe training future returns within seconds (not minutes)

## Unblocks

- Phase 2 P2-1a (`swap_dataset_live` skeleton). The spec's pause-boundary semantics (manager.py step 5 of the swap flow) now work as assumed.
- Also fixes a UX bug affecting every canopy user, not just Phase 2.

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -82,6 +82,23 @@ class InvalidCandidatePoolError(ValueError):
     """
 
 
+class TrainingInterrupted(Exception):
+    """Sentinel raised from training-loop callbacks when the user requests stop.
+
+    Pre-2026-05-10 the ``pause_training`` and ``stop_training`` REST endpoints
+    set ``_pause_event`` / ``_stop_requested`` and transitioned the FSM, but the
+    flags were never observed inside ``cascade_correlation.fit()`` — training
+    ran to natural completion regardless. The fix wires signal checks into the
+    ``_output_training_callback`` and ``_grow_iteration_callback`` hook points;
+    when ``_stop_requested`` is set, the callback raises this sentinel which
+    ``monitored_fit`` catches as a clean cancellation (FSM → STOP, status
+    Stopped, cancelled-counter incremented; not a failure).
+
+    See ``ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md`` §3.4 audit findings
+    for the original defect documentation; this fix is P2-PRE-1 of that plan.
+    """
+
+
 def _validate_candidate_pool_triple(s: int, t: int, r: int, p: int) -> Optional[str]:
     """Return ``None`` if (S, T, R, P) satisfy the §1.5 C2.1 invariant or a
     human-readable violation string otherwise.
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #252, opened 2026-05-13
feat(phase2): P2-1d — live dataset swap resize + dataset zero-pad (Issue #3)

pcalnon:
## Summary

Implements **P2-1d** of Issue #3 — the live dataset swap shrink + grow path, redesigned from the §3.6 prepend-adapter approach (abandoned 2026-05-12) to a much simpler model:

- **Grow** → in-place tensor expansion with **random-init** × `random_value_scale` (matches construction-time pattern).
- **Shrink** → **zero-pad the dataset** up to network capacity; the network never shrinks. Loss is masked to active output dims so the zero-padded target tail doesn't pull gradients toward zero.

Full design lock-in is checked in alongside the code at [`notes/PHASE_2_P2_1D_DESIGN_2026-05-13.md`](../tree/phase2/p2-1d-resize-and-pad/notes/PHASE_2_P2_1D_DESIGN_2026-05-13.md). The parent spec's §3.6 will be retired by a companion canopy PR (separate, doc-only).

### What changes

**Network class** (`cascade_correlation.py`):

- `_resize_network_for_dataset(input_size_new, output_size_new)` — grows input/output dims, returns delta descriptor. Shrink raises.
- `_grow_output_layer` — **row-insertion at index `self.input_size`** on `output_weights` (preserves the `[raw_inputs | hidden_outputs]` layout consumed by `forward()` at line 1577); column-append for output dim; matching element-append on `output_bias`.
- `_grow_hidden_units_for_input` — mirrors the row-insertion on each hidden unit's weight vector. Hidden-unit biases (scalars) untouched per cascade-correlation contract.
- New `self.active_output_dim` attribute; `train_output_layer` slices `output[:, :active_output_dim]` and `y[:, :active_output_dim]` before computing loss when active is less than full output_size.
- `calculate_residual_error` zero-fills residual columns past `active_output_dim` so multiprocess candidate training never correlates against zero-padded target signal (avoids threading masking through every candidate worker).

**Lifecycle layer** (`api/lifecycle/manager.py`):

- `_pad_dataset_for_network(x, y, val_x, val_y)` — zero-pads tensors up to network capacity; returns pre-pad active dims. Raises `ValueError` if dataset exceeds capacity.
- `swap_dataset_live` step 8 rewired: resize for grow, pad for shrink, set `network.active_output_dim`. Mixed swaps (input grows + output shrinks) compose naturally; sides are independent.
[...]

PR #251, opened 2026-05-12
feat(phase2): P2-1c — additive-only architecture adapter (grow input/output) (Issue #3)

pcalnon:
## Summary

Implements **P2-1c** of [`ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md`](../juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md) §3.6 / §7: the additive-only architecture adapter. When a live dataset swap changes input or output dimension, the relevant weight tensors are expanded **in place** with zero-initialised new connections, preserving all hidden units and the cascade structure unchanged.

Builds on **P2-1b** (#250) which landed the cancel mechanism + structured-log polish. With P2-1c in, `swap_dataset_live` now handles equal-dim and any **grow** combination; only shrink remains 422 (deferred to P2-1d, which needs its own design doc per §7).

### What changes

**New module** — `api/lifecycle/architecture_adapter.py`:

- `ArchChanges` frozen dataclass — maps 1:1 onto the §3.3 `arch_changes` response block; `to_response_dict()` is dict-copy-safe.
- `adapt_for_dataset_swap(network, before, after)`:
  - Equal-dim → no-op (network not touched; avoids `requires_grad` churn).
  - Grow-input → row-insertion (**not append**: existing hidden-unit rows shift down to make room).
  - Grow-output → column-append on `output_weights` + element-append on `output_bias`.
  - Mixed grow → apply both independently.
  - Any shrink (including mixed-with-grow) → `ValueError("shrink_unsupported (P2-1c): ...")`.

**Tensor mutations** (full inventory):

| Op | `output_weights` | `output_bias` | each `hidden_units[i]["weights"]` |
|---|---|---|---|
| equal-dim | untouched | untouched | untouched |
| grow-input `I_old→I_new` | `[I_old+H, O]` → `[I_new+H, O]`; rows shift down, gap zero-init | untouched | `[I_old+i]` → `[I_new+i]`; same row-insertion |
| grow-output `O_old→O_new` | `[N, O_old]` → `[N, O_new]`; cols append, zero-init | `[O_old]` → `[O_new]`; append zero | untouched |

`requires_grad_(True)` is restored on `output_weights` after every reassignment so the next `train_output_layer()` call can build an optimizer on the expanded tensor (matches the existing `_resize_output_layer_for_new_units` convention).

**Zero-init invariant** (§3.6 — the property that makes mid-training grow safe):

[...]

PR #250, opened 2026-05-12
feat(phase2): P2-1b — DELETE /v1/training/dataset/live cancel + polish (Issue #3)

pcalnon:
## Summary

Implements the **P2-1b** row of [`ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md`](../juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md) §7: a cancel mechanism for the live dataset swap, plus the structured-log / candidate-count / topology-rebroadcast polish called out by §3.7 guardrails #5 and #7.

Builds on **P2-1a** (#245) which landed the experimental-functions gate and the bare equal-dim `swap_dataset_live` skeleton, and **P2-PRE-1** (#244) which made `pause_training` / `stop_training` actually interrupt the training loop.

### What changes

**Lifecycle (`api/lifecycle/manager.py`)**

- New exceptions:
  - `NoSwapInProgressError` → 404 on `DELETE /v1/training/dataset/live`
  - `SwapCancelledError` → 200 with `{"status": "cancelled"}` on the originating `POST`
- `_swap_cancel_requested` (`threading.Event`) observed at two safe checkpoints inside `swap_dataset_live`:
  1. Post-fetch (most likely landing spot for a user cancel — fetch is the long pole)
  2. Pre-future-resubmit (last chance before the new training future goes live)
- `request_swap_cancel()` sets the flag under `_training_lock`. Stale flags are cleared at swap-start AND in `finally`, so DELETE-vs-swap races stay sane (per-swap, not sticky).
- `abandoned_candidate_pool_size` — captured **before** stopping the training future (reading after the future stops always yields zero). Reports the configured pool size when in `TrainingPhase.CANDIDATE`, zero otherwise (§3.5 Option C).
- §3.7 #5 structured INFO log on completion:
  `swap: input 2→3, output 2→2, hidden 5 preserved, candidates 8 abandoned, mode→output_training`
- §3.7 #7 topology rebroadcast already fired in P2-1a via `_broadcast_training_state(force=True)`; new unit test asserts it.

**Route (`api/routes/training.py`)**

- `DELETE /v1/training/dataset/live`: 403 (gate closed) / 404 (no swap) / 200 (cancel signal accepted)
- `POST /v1/training/dataset/live`: now translates `SwapCancelledError` → 200 with `{"status": "cancelled"}` so canopy's "Live Switch failed" toast does not fire on a user-initiated cancel.

[...]

PR #245, opened 2026-05-11
feat(phase2): P2-1a — experimental-functions gate + bare swap_dataset_live (Issue #3)

pcalnon:
## Summary

- Implements **Phase 2 P2-1a** per `notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md` §3.1/§3.2/§3.3/§3.7/§3.8 (spec lives in juniper-canopy, merged via PR #266).
- **Equal-dim swaps only.** Any input/output dim change is rejected with 422 `dim_change_unsupported`. P2-1c (grow adapter) and P2-1d (shrink-via-prepend-layers) will lift the restriction.
- Builds on **P2-PRE-1** (#244 — already merged) — without that pause/stop fix, this swap path would have blocked for minutes waiting for natural fit completion.

## What ships

**Manager surface** (`src/api/lifecycle/manager.py`):
- `SwapInProgressError` sentinel — promotes to 409.
- `_PreSwapSnapshot` value container for the §3.7 #1 rollback snapshot.
- `_experimental_functions_enabled` (env-overridable via `CASCOR_EXPERIMENTAL_FUNCTIONS_ENABLED=1`), `_current_dataset_config`, `_swap_in_progress`.
- `get_experimental_functions` / `set_experimental_functions` — server-side authority for F2.10. Default closed.
- `swap_dataset_live(**cfg)` — 16-step orchestrator. Held under `_training_lock` for the whole swap (Audit #2 confirmed read-side routes don't contend, so the spec's earlier release/reacquire pattern is unnecessary). On any failure: rollback via `_rollback_pre_swap_state`.
- `_reload_dataset` now tracks the canonical cfg so the swap response reports `before_cfg` accurately.

**REST surface**:
- `POST /v1/training/dataset/live` — exception-to-HTTP translation: 403 / 409 / 422 / 502 / 504.
- `GET /v1/admin/experimental_functions` — read gate state.
- `POST /v1/admin/experimental_functions` — toggle (server is authoritative per F2.10).

**Pydantic models**: `SwapDatasetLiveRequest` (inherits from `StageDatasetRequest`), `ExperimentalFunctionsToggleRequest`.

## Tests (24 new, all pass)

- `tests/integration/api/test_swap_dataset_live.py` (14) — lifecycle method contract: gate, dim-change rejection, rollback on fetch failure, equal-dim success response shape, auto-snap ratchet reset (§3.7 #6), rollback helper.
[...]

PR #253, opened 2026-05-14
feat(phase2): P2-2 — dataset_swap history event + serializer round-trip (Issue #3)

pcalnon:
## Summary

Implements **P2-2** of Issue #3 (cascor) — the `dataset_swap` history event called for by spec §3.9. Every successful `swap_dataset_live` now appends a structured event to `network.history["dataset_swaps"]`, and the HDF5 snapshot serializer round-trips the event list cleanly. Canopy's P2-7 timeline UI will consume these events to render swap markers without needing to derive them from logs or live state.

Post-P2-1d redesign, the topology is structurally unchanged across swaps, so the event payload reduces to the same `arch_changes` dict the §3.3 swap response already carries (plus a timestamp and snapshot-ID placeholders for P2-3 to backfill).

### Network class (`cascade_correlation.py`)

- Added `"dataset_swaps": []` as a sibling of `"hidden_units_added"` in the `self.history` dict.
- New `record_dataset_swap_event(*, before_cfg, after_cfg, arch_changes, pre_swap_snapshot_id=None, post_swap_snapshot_id=None) -> dict` method. Deep-copies the caller's nested dicts so post-call mutation can't ripple back into history.

Event payload (per spec §3.9):

```python
{
    "timestamp":              "2026-05-14T15:42:01.234Z",   # ISO-8601 UTC
    "before_cfg":             {...},                         # dataset cfg or None
    "after_cfg":              {...},                         # dataset cfg or None
    "arch_changes":           {...},                         # the §3.3 response dict
    "pre_swap_snapshot_id":   None,                          # P2-3 backfills
    "post_swap_snapshot_id":  None,                          # P2-3 backfills
}
```

### Lifecycle (`api/lifecycle/manager.py`)

- `swap_dataset_live` calls `record_dataset_swap_event` in the success branch, immediately after the topology rebroadcast and before the structured-log/return. Cancellation and rollback raise before this point, so failed swaps never record an event.
- The same `arch_changes` dict that flows into the route response also flows into the recorder — persisted events and live responses cannot drift.
- A recorder failure is caught + logged at `exception` level but does NOT abort the swap (the swap itself already succeeded; observability shouldn't break user-facing operations).

### Serializer (`snapshots/snapshot_serializer.py`)

- **Save**: each event becomes an HDF5 subgroup at `history/dataset_swaps/event_{i}` with scalar attrs + JSON-encoded string attrs for nested dicts. Empty list → no group on disk.
- **Load**: initialises `history["dataset_swaps"] = []` (backward compat for pre-P2-2 snapshots), then reads each `event_{i}` sorted by numeric suffix (HDF5 group iteration order isn't guaranteed otherwise). Malformed JSON in one event degrades that event to schema defaults with a warning, leaving other events intact.

## Test coverage

- **6 unit tests** for `record_dataset_swap_event`: schema, chronological order, deep-copy guarantee, None cfgs, explicit snapshot IDs, empty initial state.
- **7 unit tests** for serializer round-trip: empty list, single event, multi-event ordering, None cfgs encoded as JSON `"null"` and decoded back to Python `None`, populated snapshot IDs, pre-P2-2 backward compat (missing HDF5 group → empty list), corrupt arch_changes JSON degrades to default.
- **4 integration tests** on `swap_dataset_live`: success records exactly one event matching the §3.3 response, cancelled swap records 0, fetch-failure records 0, equal-dim swap still records (dataset metadata changed even if dims didn't).

## Follow-ups (deferred, captured in `notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md`)

- **Follow-up A**: real-time WebSocket broadcast of `dataset_swap` events. Asymmetric to add for just this event type — the right design conversation is "should all history events broadcast?", bigger than P2-2.
- **Follow-up B**: `GET /v1/history/dataset_swaps` route for canopy P2-7 to fetch the event list without pulling a full snapshot. Shape can wait until P2-7 is being written.

Both are small (~50–150 LOC each), independent, and can land any time post-P2-2.

## Verification

- [x] 3562 tests pass, 15 skipped (+17 net new tests, no regressions)
- [x] 44 focused tests on the P2-2 surface pass (unit + serializer + integration)
- [x] Pause/stop interrupt regression tests pass (P2-PRE-1 healthy)

## Refs

- Spec: `juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md` §3.3 / §3.9 / §7 P2-2 row
- Follow-up design notes: `notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md` (in this PR)
- Series: P2-PRE-1 (#244) → P2-1a (#245) → P2-1b (#250) → P2-1c (#251) → P2-1d (#252) → **P2-2 (this PR)** → P2-3

### DIFF
--- a/notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md
+++ b/notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md
@@ -1,0 +1,83 @@
+# P2-2 Follow-ups: real-time broadcast + history fetch route
+
+**Status**: Captured 2026-05-14 during the P2-2 design check-in. Not yet implemented.
+**Parent PR**: P2-2 (`phase2/p2-2-dataset-swap-history`) — landed only the history persistence + serializer round-trip.
+
+This doc records two ideas surfaced during the P2-2 scoping discussion that were intentionally deferred to keep the P2-2 PR focused on persistence. Each is a small, independent follow-up that can land any time after P2-2 merges.
+
+---
+
+## Follow-up A: real-time WebSocket broadcast of `dataset_swap` events
+
+### What
+
+Push the `dataset_swap` event over the existing cascor WebSocket immediately on swap completion, so canopy can render a timeline marker without polling or fetching a snapshot.
+
+### Why deferred from P2-2
+
+History events are currently HDF5-only — no other history surface broadcasts in real time today (the agent investigation 2026-05-14 confirmed `train_loss` / `value_loss` / `train_accuracy` / `value_accuracy` / `hidden_units_added` all persist to HDF5 but do not broadcast). Adding broadcast for `dataset_swap` alone would create asymmetry — the right design conversation is "should all history events broadcast?", which is bigger than P2-2.
+
+Until this follows up, canopy P2-7 will pick up swap events via the snapshot/history GET path.
+
+### Sketch
+
+- In `swap_dataset_live` step 16 (after `record_dataset_swap_event` returns the event dict, in cascor `manager.py`), call `self._ws_manager.broadcast(create_dataset_swap_message(event))`.
+- New helper `create_dataset_swap_message(event)` in the same module as `create_event_message` / `create_cascade_add_message` (cascade_correlation.py around line 1281 per the agent's investigation — confirm before implementing).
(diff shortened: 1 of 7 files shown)


pcalnon (commented):
approved

PR #254, opened 2026-05-14
feat(phase2): P2-3 — pre/post-swap auto-snapshot + ID threading (Issue #3)

pcalnon:
## Summary

Implements **P2-3** of Issue #3 — the snapshot infrastructure half. The replay-engine rework half is **deferred** per a design discovery (`_ReplaySession` is a passive time-indexed metric playback engine, not event-sourced), with a follow-up design doc + trigger conditions captured in `notes/PHASE_2_P2_3_FOLLOWUP_REPLAY_REWORK_2026-05-14.md` so the rework can be picked up if/when canopy-orchestrated transitions prove insufficient.

The §3.9 *"instantaneous transformation"* requirement is satisfied in user-visible terms by canopy P2-7 restarting replay from `post_swap_snapshot_id` when the timeline cursor crosses a swap boundary. The mechanism is orchestration-layer; the user sees the topology change appear instantly at the marker.

### `save_snapshot` collision fix

Snapshot IDs were second-resolution (`snapshot_YYYYMMDDTHHMMSSZ`) and **silently overwrote** when two snapshots fired within the same wall-clock second — exactly the pre/post-swap pair P2-3 creates. We now check for an existing file and append `_2`, `_3`, … to the ID until a free slot is found. The no-collision path is byte-identical to the pre-P2-3 behaviour, so canopy's existing snapshot-ID parser keeps working unchanged.

A prior test in `test_can_015h_roundtrip.py` carried a `time.sleep(1.1)` workaround for this exact issue with a "tracked as a separate concern" comment — that workaround is now obsolete (kept in place; out of scope to remove).

### Pre-swap auto-snap in `swap_dataset_live`

Fires after the rollback snapshot capture (step 4) but **before any network mutation**, so the on-disk snapshot captures the genuine pre-swap state. Failure-tolerant: a snapshot write failure logs at exception level but does NOT abort the swap. The pre-swap snapshot is **kept on disk even when the swap is later cancelled / rolled back** — it's a valid state record at the moment the user attempted the swap, and cleanup would just add race conditions.

### Post-swap auto-snap in `swap_dataset_live`

Fires after the topology rebroadcast (step 14), before `record_dataset_swap_event`. Same failure tolerance.

### ID threading into `record_dataset_swap_event`

`pre_swap_snapshot_id` and `post_swap_snapshot_id` (the placeholder fields P2-2 left as `None`) are now populated from the captured snapshot IDs. Either may still be `None` if its snapshot write failed — the event still records, just without that affordance for canopy P2-7.

### Snapshot descriptions encode swap metadata

`"pre-swap before dataset_swap (from spirals)"` / `"post-swap after dataset_swap (to moons)"` so the Snapshots tab is human-browsable without joining against the history event metadata.

### Replay rework follow-up doc

`notes/PHASE_2_P2_3_FOLLOWUP_REPLAY_REWORK_2026-05-14.md` documents:

- Why `_ReplaySession` can't host a mid-playback `dataset_swap` handler today (passive time-indexed playback, no event dispatch).
- What P2-3 actually ships and how canopy P2-7 will use the snapshot IDs to render a "continue past swap" affordance.
- A specification for the event-sourced replay rework (event queue, dispatch table, per-event handlers, backward-compat strategy) with explicit trigger conditions to revisit: snapshot-reload feels jarring, multi-swap latency, animated topology transitions, mid-replay editing, replay-only mode.

## Test plan

- [x] **4 unit tests** for `save_snapshot` collision: no-collision keeps legacy format; same-second pair produces `_2` suffix; three calls cascade through `_2` and `_3`; suffixed files are valid HDF5.
- [x] **4 integration tests** for `swap_dataset_live`: success takes pre AND post snapshots in order with metadata-encoded descriptions; pre-swap captures pre-resize dims (ordering invariant); pre-swap snap kept on cancellation (event not recorded but snapshot persists); pre-swap snap failure does not abort the swap.
- [x] **1 P2-2 assertion update**: `pre_swap_snapshot_id` is no longer `None` — now asserts `startswith("snapshot_")`. Other P2-2 unit/integration tests unchanged.
- [x] **3572 tests pass**, 15 skipped (+10 net new tests, no regressions)
- [x] 52 focused tests on the P2-3 surface + P2-2 regression pass

## Out of scope (per design doc)

- Event-sourced replay engine + per-event handlers. Trigger conditions documented in the follow-up doc.
- Animated topology transitions in canopy timeline UI.
- Mid-replay editing / "what-if" UX.

## Refs

- Spec: `juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md` §3.3 / §3.9 / §7 P2-3 row
- Follow-up design: `notes/PHASE_2_P2_3_FOLLOWUP_REPLAY_REWORK_2026-05-14.md` (in this PR)
- Series: P2-PRE-1 (#244) → P2-1a (#245) → P2-1b (#250) → P2-1c (#251) → P2-1d (#252) → P2-2 (#253) → **P2-3 (this PR)** → P2-7 (canopy)

### DIFF
--- a/notes/PHASE_2_P2_3_FOLLOWUP_REPLAY_REWORK_2026-05-14.md
+++ b/notes/PHASE_2_P2_3_FOLLOWUP_REPLAY_REWORK_2026-05-14.md
@@ -1,0 +1,105 @@
+# P2-3 follow-up: event-sourced replay rework
+
+**Status**: Captured 2026-05-14 during P2-3 implementation. Not yet scheduled.
+**Parent PR**: P2-3 (`phase2/p2-3-swap-snapshots-replay`).
+
+P2-3's parent spec ([`juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md`](../../juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md) §3.9 / §7) calls for a *"Replay reconstruction handler (instantaneous transformation, per §8 Answer 3)"* — implying that when replay encounters a `dataset_swap` event mid-playback, a handler fires and transforms the network's dims instantaneously.
+
+The 2026-05-14 P2-3 implementation investigation revealed that **`_ReplaySession` is a passive time-indexed metric playback engine**, not an event-sourced replay engine. The "mid-playback handler" pattern cannot be implemented on top of the current engine without substantial architectural rework. This document captures (a) why P2-3 shipped without the mid-playback handler, (b) the canopy-orchestrated alternative P2-3 enables, and (c) the rework spec to revisit if the alternative turns out insufficient.
+
+---
+
+## 1. Current replay engine (`_ReplaySession`)
+
+Defined at `src/api/lifecycle/manager.py:766`.
+
+Flow on `start_replay(snapshot_id)`:
+
+1. Load the target HDF5 snapshot via `_load_snapshot_to_network()`. The deserializer reconstructs the entire network state — weights, topology, history arrays — exactly as it was at the moment the snapshot was taken.
+2. The replay session captures `network.history` (metric arrays) and `network.weight_history` (CAN-015g per-sample weights).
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #255, opened 2026-05-15
feat(phase2): P2-2 follow-ups A+B — WS broadcast + GET history route (Issue #3)

pcalnon:
## Summary

Lands both P2-2 follow-ups documented in [`notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md`](../tree/phase2/p2-2-followups-broadcast-and-fetch-route/notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md) as one PR — they share the *dataset_swap event observability* theme and the same files, so bundling avoids two stub PRs of ~150 LOC each.

After this lands, canopy P2-7's timeline UI has two ways to consume `dataset_swap` events: poll the new GET route (good for cold loads / catch-up) or subscribe to the existing WebSocket (good for real-time rendering of swaps that fire while the user is watching).

### Lifecycle

New `TrainingLifecycleManager.get_dataset_swap_events(since=None)`:

- Reads from `network.history["dataset_swaps"]` (P2-2's persistence surface).
- Optional ISO-8601 `since` filter for polling clients — returns events with timestamp **strictly greater** than `since`, so a poller can pass its last-seen timestamp and get only new events.
- Skips events without a timestamp when `since` is set (defensive against malformed snapshots).
- **Deep-copies** each event so callers iterating the result can mutate nested dicts (`arch_changes`, `appended_nodes`) without corrupting persisted history. Matches the recorder's own deep-copy guarantee.

### Route

New `src/api/routes/history.py` with `GET /v1/history/dataset_swaps`. Thin shim — passes `since` through verbatim. Registered in `app.py` after the admin router.

## Follow-up A — WebSocket broadcast of dataset_swap

`swap_dataset_live` now captures the event dict returned by `record_dataset_swap_event` and broadcasts it via the existing generic `create_event_message({"event": "dataset_swap", "swap": event_dict})`.

This avoids needing a new envelope class in the external `juniper_cascor_protocol` package — the `event`-discriminator pattern is already used by `training_complete` (`manager.py:1342`).

The broadcast is **failure-tolerant**: a WS error logs at exception level but does NOT abort the swap. The swap has already succeeded; observability failures must not roll back user data. No-op when `_ws_manager` is None.

## Test plan

- [x] **5 unit tests** for `get_dataset_swap_events`: empty when no network, empty when history has no swaps, returns all without filter, strictly-after with `since`, skips timestamp-less when `since` set, deep-copy guarantee
- [x] **4 route tests**: empty list, chronological order, `since` query param passed verbatim, 503 when lifecycle missing
- [x] **4 integration tests** for the WS broadcast: success broadcasts exactly one event with the right envelope shape, cancellation does NOT broadcast, no-op when `_ws_manager` is None, broadcast failure does not abort the swap (selective side_effect targeting only the dataset_swap envelope)
- [x] **3586 tests pass**, 15 skipped (+14 net, no regressions)

## Wire format

Broadcast envelope (consumed by canopy WS clients):

```json
{
  "type": "event",
  "timestamp": 1747353600.123,
  "data": {
    "event": "dataset_swap",
    "swap": {
      "timestamp": "2026-05-14T19:55:00+00:00",
      "before_cfg": {"dataset_type": "spirals"},
      "after_cfg":  {"dataset_type": "moons"},
      "arch_changes": {...},
      "pre_swap_snapshot_id":  "snapshot_...",
      "post_swap_snapshot_id": "snapshot_..._2"
    }
  }
}
```

Same `swap` shape canopy gets from the GET route — push and pull paths converge on a single event schema.

## Refs

- Follow-up doc: `notes/PHASE_2_P2_2_FOLLOWUPS_2026-05-14.md` (in cascor main)
- Parent spec: `juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md`
- Series: P2-PRE-1 (#244) → P2-1a (#245) → P2-1b (#250) → P2-1c (#251) → P2-1d (#252) → P2-2 (#253) → P2-3 (#254) → **P2-2 follow-ups A+B (this PR)** → canopy track

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -17,7 +17,7 @@ from api.lifecycle.manager import TrainingLifecycleManager
 from api.middleware import RequestBodyLimitMiddleware, SecurityHeadersMiddleware, SecurityMiddleware
 from api.models.common import error_response
 from api.observability import PrometheusMiddleware, RequestIdMiddleware, configure_logging, configure_sentry, get_prometheus_app, set_build_info
-from api.routes import admin, dataset, decision_boundary, health, metrics, network, snapshots, training, workers
+from api.routes import admin, dataset, decision_boundary, health, history, metrics, network, snapshots, training, workers
 from api.secrets import get_secret
 from api.security import APIKeyAuth, RateLimiter
 from api.settings import Settings, get_settings
@@ -467,6 +467,7 @@ def create_app(settings: Settings | None = None) -> FastAPI:
     app.include_router(snapshots.router, prefix="/v1")
     app.include_router(workers.router, prefix="/v1")
     app.include_router(admin.router, prefix="/v1")  # Phase 2 P2-1a — experimental-functions gate
+    app.include_router(history.router, prefix="/v1")  # Phase 2 P2-2 Follow-up B — dataset_swap event fetch
 
     # WebSocket Routes
     app.websocket("/ws/training")(training_stream_handler)
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -2324,6 +2324,42 @@ class TrainingLifecycleManager:
         cfg = self._pending_dataset_config
         return dict(cfg) if cfg else None
 
+    # P2-2 Follow-up B (Issue #3): canopy P2-7 timeline UI reads dataset_swap
+    # events via this lifecycle method without pulling a full snapshot.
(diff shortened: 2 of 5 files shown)


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #258, opened 2026-05-18
fix(cascade_correlation): restore missing _ACTIVATION_FUNCTION_DEFAULT import

pcalnon:
## Summary

- `cascade_correlation.py` references `_CASCADE_CORRELATION_NETWORK_ACTIVATION_FUNCTION_DEFAULT` at four sites (L1037, L1039, L1448, L3388) but the symbol was dropped from the import list pulled out of `cascor_constants.constants`.
- Every `CascadeCorrelationNetwork(...)` construction raised `NameError` at the first reference inside `_init_activation_function`, breaking fixture setup for any test that instantiates a network.
- This PR adds the one missing import line. No behavioural change beyond restoring the previously intended default-activation lookup path.

## Test plan

- [x] `PYTHONPATH=src python -c "from cascade_correlation.cascade_correlation import CascadeCorrelationNetwork; CascadeCorrelationNetwork(config=...)"` succeeds locally on `JuniperCascor1`.
- [ ] CI green on the fix branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #259, opened 2026-05-18
feat(phase2): P2-7 follow-up — GET /v1/snapshots/{id}/history/dataset_swaps (Issue #3)

pcalnon:
## Summary

Surfaces a stored snapshot's own `history/dataset_swaps` event list so canopy's Replay timeline can render swap markers tied to the **loaded snapshot's** history (parent spec §4.4 full flavor), separate from the live event feed exposed by `GET /v1/history/dataset_swaps` (P2-2 follow-up B).

PR #277 (canopy P2-7) explicitly deferred this surface — its timeline reads only the live event feed, semantically equivalent for most flows but not faithful to §4.4 when a user is reviewing a stored snapshot whose swap history might differ from the live feed.

## New surface

```
GET /v1/snapshots/{snapshot_id}/history/dataset_swaps
→ 200 { "data": { "events": [ ... §3.9 schema ... ] } }
→ 200 { "data": { "events": [] } }   # no swaps in this snapshot
→ 404 { "detail": "Snapshot '...' not found" }
```

## Why a separate endpoint (not an inline field on `GET /v1/snapshots/{id}`)

`get_snapshot` is currently a `stat()`-only call; list/poll clients rely on it staying cheap. Opening the HDF5 to extract `history/dataset_swaps` would make every metadata fetch slower. The sibling endpoint defers that cost to the moments when canopy actually needs the markers — i.e. when a snapshot is loaded for replay.

## Layering

- `snapshot_serializer.py`: pulls the existing in-place dataset_swap decode loop into a static `_decode_dataset_swap_events_from_group` used by both `_load_training_history` (full network restore) and a new public `read_dataset_swap_events(filepath)` (header-only HDF5 open, no network restore). Behaviour-preserving refactor; the loader path is byte-identical to before.
- `api/lifecycle/manager.py`: new `get_snapshot_dataset_swaps(snapshot_id)` returns `None` on missing snapshot (404 cue), `[]` for snapshots without swaps.
- `api/routes/snapshots.py`: new sibling route to `GET /v1/snapshots/{id}`.

## Stacks on

- **PR #258** (`fix/cascade-correlation-default-activation-import`) — restores a missing `cascor_constants` import that blocks every network-constructing test. Required for this PR's tests to run at all.

The two stacked commits at the bottom of this PR's diff (`c44041a` and `c7f1f44`) are the cherry-picks of PR #258's contents; they disappear cleanly on rebase after #258 merges.

## Test plan

- [x] **5 serializer-level tests** (`read_dataset_swap_events`): without-load, empty list, absent history group, chronology parity vs the loader, corruption-tolerance parity vs the loader.
- [x] **4 lifecycle tests** (`get_snapshot_dataset_swaps`): 404 cue, empty list, chronological order, schema parity with `get_dataset_swap_events` (P2-2 follow-up B).
- [x] **5 route tests**: 404 / 200-empty / chronological-200 / invalid-id / lifecycle-not-initialised → 503.
- [x] 82 tests pass across the touched serializer + dataset_swap suites; no regressions.
- [x] `pre-commit run` clean on the changed files (black, isort, flake8, mypy, bandit, async-route audit).

## Refs

- Spec: `juniper-canopy/notes/ISSUE_3_PHASE_2_LIVE_DATASET_SWAP_2026-05-09.md` §3.9 / §4.4
- Deferred bullet: PR #277 body — "Snapshot-history fetch for replay-context markers"


pcalnon (commented):
approved

PR #270, opened 2026-05-19
fix(cascade_correlation): BUG-CC-09 — re-validate max_epochs after output_epochs fall-back

pcalnon:
## Summary

`fit()` resolved a `None` `max_epochs` argument from `self.output_epochs` **after** the only `_validate_positive_integer` call, so a misconfigured `self.output_epochs` (0 or negative) bypassed validation and reached `train_output_layer` with a non-positive value. The training loop is `for epoch in range(epochs):` — `range(0)` / `range(<0)` yield no iterations, so the body never runs, weights never update, and the returned `final_loss` reflects the *unchanged* forward pass. Silent no-op training (v7 roadmap §5.1 BUG-CC-09).

## Fix

Re-validate `max_epochs` after the `(max_epochs, self.output_epochs)[max_epochs is None]` resolution by reusing the existing `_validate_positive_integer` helper. 3-line change.

## Why this approach, not the roadmap's silent-skip

The v7 roadmap suggested `if max_epochs <= 0: train_loss = float("inf")`. That would mask the misconfiguration. Hard-failing via the existing validator surfaces the config error at the call site rather than producing a "trained" network that never saw a gradient update.

## Test plan

- [x] New `test_fit_rejects_zero_output_epochs_misconfig` pins the ValidationError for `self.output_epochs == 0`.
- [x] New `test_fit_rejects_negative_output_epochs_misconfig` covers negative values.
- [x] 5/5 `TestFitMethod` tests pass.
- [x] `pre-commit run` clean.

## Refs

- v7 roadmap §5.1 — BUG-CC-09

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -1770,6 +1770,14 @@ class CascadeCorrelationNetwork:
         self.logger.trace("CascadeCorrelationNetwork: fit: Starting initial training of the output layer.")
         self.logger.info("CascadeCorrelationNetwork: fit: Initial training of output layer")
         max_epochs = (max_epochs, self.output_epochs)[max_epochs is None]
+        # BUG-CC-09: re-validate after the ``self.output_epochs`` fall-back —
+        # the earlier check only ran when ``max_epochs`` came in non-None,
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #267, opened 2026-05-19
fix(ci): unbreak pre-commit on main — exclude conda explicit-format yaml + delete stale snapshots

pcalnon:
## Summary

Parallel to juniper-data PR #117 — same root cause and fix shape.

CI pre-commit has been failing on main because \`conf/conda_environment_ci.yaml\` is generated by \`scripts/generate_dep_docs.sh\` using conda's explicit-list format (YAML preamble + bare package-URL list). Bare URLs after the \`dependencies:\` key make the file invalid as pure YAML. Both \`check-yaml\` and \`yamllint\` fail on every PR.

Five stale timestamped backups from 2026-05-18 (all with colons in the filename) had been committed to main; they suffer from the same parsing issue.

Discovered while investigating the branch-protection refresh — required\\_status\\_checks for Pre-commit would be impossible to satisfy until this is fixed.

## Requirements

- References JR-CAS-CI-* — CI hygiene; no individual JR-* entry being closed.

## What changed

| File | Change |
|------|--------|
| \`.pre-commit-config.yaml\` | Add \`exclude:\` regex to both \`check-yaml\` and \`yamllint\` hooks: \`^conf/conda_environment_ci(_[0-9].*)?\\.yaml$\`. |
| \`.yamllint.yaml\` | Add the same paths to the \`ignore:\` list (documentation; no-op under pre-commit's \`-d relaxed\` override). |
| \`.gitignore\` | Block future timestamped snapshots from being committed; keep canonical files tracked via \`!\` unignore. |
| \`conf/conda_environment_ci_2026-05-18_17:17:17.yaml\` | **DELETED** |
| \`conf/conda_environment_ci_2026-05-18_17:21:29.yaml\` | **DELETED** |
| \`conf/conda_environment_ci_2026-05-18_17:21:50.yaml\` | **DELETED** |
| \`conf/requirements_ci_2026-05-18_17:21:29.txt\` | **DELETED** |
| \`conf/requirements_ci_2026-05-18_17:21:50.txt\` | **DELETED** |

## Follow-ups (NOT in this PR)

1. The \`.yaml\` extension on \`conf/conda_environment_ci.yaml\` is a misnomer — conda's explicit-list output belongs in a \`.lock\` or \`.txt\` file. Renaming touches references in docs and other scripts; tracked for a separate PR.
[...]

PR #272, opened 2026-05-19
fix(ci): restore scripts/generate_dep_docs.sh as a real file (Dependency Documentation)

pcalnon:
## Summary

Repair the dangling symlink left behind by PR #268. CI's Dependency Documentation step has been failing on \`main\` because \`scripts/generate_dep_docs.sh\` is a mode-120000 symlink whose target (\`../util/generate_dep_docs.bash\`) was deleted as part of the same PR. Fresh CI checkouts hit \`bash: scripts/generate_dep_docs.sh: No such file or directory\`.

This is the follow-up I flagged in PR #271's body. It was anticipated, but reverted to a real file here rather than restoring the symlink target — fewer indirection layers, matches the convention in juniper-data-client / juniper-data / juniper-canopy / juniper-cascor-worker / juniper-cascor-client (all of which keep \`scripts/generate_dep_docs.sh\` as a real file).

## Requirements

- References JR-CAS-CI-* — CI hygiene.

## What changed

| File | Change |
|------|--------|
| \`scripts/generate_dep_docs.sh\` | Mode 120000 → 100755. Symlink target swapped for the real script content (the dash-timestamp variant already in canopy / data-client / ml). |

The mode change is the substantive change; the file content matches what PR #268 *intended* to land but the cp-follows-symlink semantics prevented from landing.

## What this does NOT fix (deferred)

- The script still uses \`conda list --explicit\` which produces invalid-YAML (the YAML preamble + \`@EXPLICIT\` URL list issue). The \`yamllint\` excludes added in PR #267 stay needed. A separate PR can port juniper-canopy's \`conda env export --no-builds\` upgrade, which would let those excludes be removed.

## Test plan

- [x] \`pre-commit run --files scripts/generate_dep_docs.sh\` — passes (shellcheck OK).
- [x] Fresh-checkout verification: \`file scripts/generate_dep_docs.sh\` now reports "Bourne-Again shell script" (was: "broken symbolic link").
- [x] Dependency Documentation should turn green on CI after this lands.

### DIFF
--- a/scripts/generate_dep_docs.sh
+++ b/scripts/generate_dep_docs.sh
@@ -1,1 +1,129 @@
-../util/generate_dep_docs.bash
+#!/usr/bin/env bash
+#####################################################################################################################################################################################################
+# Project:       Juniper
+# File Name:     generate_dep_docs.sh
+# Author:        Paul Calnon
+#
+# License:       MIT License
+# Copyright:     Copyright (c) 2024-2026 Paul Calnon
+#
+# Description:
+#    Generates dependency documentation files for CI/CD.
+#    - requirements_ci.txt (pip freeze with header)
+#    - conda_environment_ci.yaml (conda list --explicit with header)
+#    - Preserves existing files with timestamped backups
+#
+# Usage:
+#    bash scripts/generate_dep_docs.sh
+#
+#####################################################################################################################################################################################################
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #278, opened 2026-05-20
fix(utils): CLN-CC-10 — guard dill import + add [debug] extra

pcalnon:
## Summary

- Guard `import dill` inside `utils.check_object_pickleability` so it raises a clean `ImportError` pointing at `pip install juniper-cascor[debug]` instead of a confusing stack-trace when dill is missing.
- Add a focused `debug = ["dill>=0.3.8"]` optional extra in `pyproject.toml` (also added to `[all]`). `dill` is still pinned in `[test]` since the test stack uses it; the new extra exists so users can opt in to the debugging helper without pulling the full test toolchain.
- Drop the misleading `# TODO: This function is borked` comment — once the import is guarded, the function is functional whenever dill is available, which is exactly the design that's been in place since `dill>=0.3.6` landed in `[test]`.

Roadmap reference: §6 CLN-CC-10 in `juniper-ml/notes/JUNIPER_OUTSTANDING_DEVELOPMENT_ITEMS_V7_IMPLEMENTATION_ROADMAP.md`.

## Behavior

| Environment | Before | After |
|-------------|--------|-------|
| `[test]` / `[debug]` installed | works | works (unchanged) |
| dill missing | raises `ModuleNotFoundError: No module named 'dill'` | raises `ImportError: check_object_pickleability requires the 'dill' package. Install with: pip install juniper-cascor[debug]` |

`check_object_pickleability` is internal debugging surface used during development to diagnose multiprocessing pickle failures. Only test files call it; production code does not.

## Test plan

- [x] `pytest src/tests/unit/test_utils_extended.py -k pickleability -v` → 11 passed
- [x] `pytest src/tests/unit/test_utils_extended.py src/tests/unit/test_utils_coverage.py -q` → 78 passed (no regressions)
- [x] Manual ImportError-path check via `sys.modules['dill'] = None` → receives the expected actionable message
- [x] `pre-commit run --files pyproject.toml src/utils/utils.py` → all hooks pass
- [x] `pip install -e ".[debug]"` resolves successfully in JuniperCascor1


pcalnon (commented):
approved

PR #290, opened 2026-05-21
chore(ci): Wave 4 — delete inline scripts/generate_dep_docs.sh (replaced by juniper-ci-tools)

pcalnon:
## Summary

Wave 4 of the dep-docs PyPI migration plan
(`juniper-ml:notes/JUNIPER_CI_TOOLS_PYPI_MIGRATION_PLAN_2026-05-20.md`).

Deletes the inline `scripts/generate_dep_docs.sh` that Wave 2
replaced with the `juniper-generate-dep-docs` console script from
the PyPI-published `juniper-ci-tools>=0.1.0,<0.2.0` package.

The CI `Generate Dependency Documentation` step has been invoking
the console script since Wave 2 merged on 2026-05-21 (verified green
across all 7 consumer repos), so the inline copy is now dead code
and removing it eliminates the drift surface that motivated the
migration in the first place (three coexisting copies of the same
script with a YAML-corrupting bug fix in only one).

## Verification

- `bash scripts/generate_dep_docs.sh` no longer appears in
  `.github/workflows/ci.yml` (Wave 2, already merged).
- `pip install "juniper-ci-tools>=0.1.0,<0.2.0"` +
  `juniper-generate-dep-docs` is the canonical invocation
  going forward.
- No other in-tree code path invokes the deleted script
  (`git grep "bash scripts/generate_dep_docs.sh"` returns no callers
  after this PR).

## Rollback

If a regression surfaces, revert this commit; the Wave 2
ci.yml change does not depend on the inline script being
present. A revert alone restores the script.


pcalnon (commented):
approved

PR #297, opened 2026-05-21
fix(CFG-04): consolidate JUNIPER_DATA_URL lookup onto Settings field

pcalnon:
## Summary

CFG-04 consolidates cascor's `JUNIPER_DATA_URL` env-var lookup onto a new `Settings.juniper_data_url: str | None` pydantic field in `src/api/settings.py`. The 8 historical raw `os.environ.get` / `os.getenv` call sites — spread across 6 files — now read from the field instead.

## Context

Historically the same env var was duplicated across 8 raw lookups:

| File | Sites |
|---|---|
| `src/api/app.py` | 3 (`create_app()`, `_auto_start_training`, `_auto_start_canopy`) |
| `src/api/routes/health.py` | 1 (readiness probe) |
| `src/api/lifecycle/manager.py` | 1 (pending-dataset reload) |
| `src/main.py` | 1 (pre-flight check) |
| `src/spiral_problem/spiral_problem.py` | 1 (`generate_n_spiral_dataset`) |
| `src/spiral_problem/data_provider.py` | 1 (`SpiralDataProvider.__init__`) |

Duplicate raw lookups make the contract invisible to operators (it doesn't surface in `Settings.model_dump()` or generated config docs) and force re-discovery every time the field needs to be validated, defaulted, or aliased.

## Design

The field is exposed via:

```python
juniper_data_url: str | None = Field(
    default=None,
    validation_alias=AliasChoices(
        "juniper_data_url",
        "JUNIPER_DATA_URL",                  # canonical, ecosystem-wide
        "JUNIPER_CASCOR_JUNIPER_DATA_URL",   # per-service override
    ),
)
```

`AliasChoices` precedence is left-to-right, so the canonical unprefixed `JUNIPER_DATA_URL` (the shared name also used by juniper-data and juniper-canopy) takes precedence over the per-service prefixed alias. Deployments that pin a global `JUNIPER_DATA_URL` are not silently overridden by a leftover prefixed value.

### Not a deprecation migration

Unlike CFG-03 (`SENTRY_SDK_DSN` → `JUNIPER_CASCOR_SENTRY_DSN`) and CFG-05 (`CASCOR_LOG_LEVEL` → `JUNIPER_CASCOR_LOG_LEVEL`):

- **No env-var rename** — `JUNIPER_DATA_URL` is the canonical and indefinitely-supported name.
- **No `DeprecationWarning`** — the prefixed alias is **additive**, giving operators a way to apply a cascor-only override without polluting juniper-data / juniper-canopy on the same host.
- **No "removed in a future release" timeline** — both names stay valid.

The CHANGELOG entry sits under `### Changed`, not `### Deprecated`, to reflect this.

### Behavior is byte-for-byte unchanged

- **Default `None`** — same as `os.environ.get("JUNIPER_DATA_URL")` when unset.
- **4 fallback sites** (`app.py` x3, `lifecycle/manager.py`) apply the legacy `http://localhost:8100` default explicitly via `settings.juniper_data_url or _PROJECT_API_JUNIPER_DATA_URL_DEFAULT`.
- **4 required sites** (`main.py` pre-flight, `SpiralProblem.generate_n_spiral_dataset`, `SpiralDataProvider.__init__` / `validate_configuration` / `get_spiral_dataset`, readiness probe) check for `None`/empty and short-circuit identically.
- **Runtime / per-request sites** construct a fresh `Settings()` rather than going through the cached `get_settings()`. Pydantic construction is sub-millisecond and none of these are hot paths; this preserves use-time env-read semantics so tests that `patch.dict(os.environ, {...})` between calls pick up changes without cache plumbing.
- **Startup sites** use the `settings` instance already in scope (`create_app`, `_auto_start_training`, `_auto_start_canopy`).

## Test impact

One existing test required an update:

- `test_api_app_coverage_deep.py::test_auto_start_uses_environment_variables` previously pre-constructed `Settings(...)` then patched `JUNIPER_DATA_URL` afterwards — relying on `_auto_start_training` reading the env at call time. Now it passes `juniper_data_url=` directly to `Settings(...)`, which is the post-CFG-04 equivalent and stays decoupled from ambient process env.

New 7-case regression suite at `src/tests/unit/test_cfg_04_juniper_data_url_settings.py` pins the contract:

1. Default `None` when neither env var is set
2. `JUNIPER_DATA_URL` (canonical) populates the field
3. `JUNIPER_CASCOR_JUNIPER_DATA_URL` (prefixed) populates the field
4. Both set → canonical wins (left-to-right `AliasChoices`)
5. `Settings(juniper_data_url=...)` kwarg overrides env
6. Empty-string env preserved as-is (matches pre-CFG-04 `os.environ.get` semantics)
7. Annotation contract `str | None`

## Validation

- ✅ CFG-04 regression suite: 7 / 7 passing
- ✅ All affected modules' existing tests: 140 / 140 passing
- ✅ Full unit test suite: **3592 passed**, 15 skipped (slow tests)
- ✅ Pre-commit hooks: black, isort, flake8, mypy, bandit, async-route-audit, doc-links — all pass

## Roadmap

Tracks CFG-04 in the v7 outstanding-development roadmap.


pcalnon (commented):
approved

PR #303, opened 2026-05-24
fix(ci): drop --cov= from scheduled slow-unit-tests step (unblocks nightly)

pcalnon:
## Summary

The **Scheduled Long Tests** workflow has been failing nightly for **27 of the last 30 runs** (last green 2026-05-03; consecutive red since 2026-05-04). The failure is **entirely in `slow-unit-tests`**; the other three jobs (`slow-integration-tests`, `long-running-tests`, `performance-benchmarks`) pass. This PR drops `--cov=src` from the slow-unit-tests step so the nightly stops red-flagging.

## Root cause

The step ran:

```yaml
python -m pytest \
  -m "unit and slow" src/tests/unit \
  ... \
  --cov=src \
  --cov-report=term-missing \
  --cov-report=xml:reports/coverage-slow.xml
```

Only ~10 tests carry `@pytest.mark.slow`. Coverage of src/ from those 10 tests is **33.75%**. `pyproject.toml` ships:

```toml
[tool.coverage.report]
fail_under = 80
```

So `pytest --cov=src` correctly exits 1 with `FAIL Required test coverage of 80.0% not reached. Total coverage: 33.75%`, even though every test passed (junit XML confirms `failures="0" errors="0"`). Job's been impossible to satisfy since juniper-cascor PR #186 (merged 2026-05-03) created `src/tests/reports/` so pytest-cov could actually write data — before that, the same gate failed with the more cryptic "No data to report".

The full-suite coverage gate already runs in `ci.yml`'s `unit-tests` job over the proper `unit and not slow` test set. Running it again on a 10-test subset was always going to fail.

## Changes

- `.github/workflows/scheduled-tests.yml` — drop `--cov=src`, `--cov-report=term-missing`, and `--cov-report=xml:reports/coverage-slow.xml` from the `slow-unit-tests` step; drop the now-unused `reports/coverage-slow.xml` from upload-artifact path. JUnit XML still uploads. Inline comment explains why.
- `src/tests/unit/test_workflow_coverage_gate.py` (new, 133 lines) — workflow-lint regression test. Scans `.github/workflows/*.yml`, stitches multi-line pytest invocations across backslash continuations, and asserts that any pytest call combining `--cov=` with a partial marker filter (`slow` / `long` / `performance`) also passes `--cov-fail-under=0`, `--no-cov-fail-under`, or `--no-cov`. Three test methods cover: workflows dir exists, scanner finds pytest invocations, no offending pattern is present.

## Test plan

- [x] Local: `pytest src/tests/unit/test_workflow_coverage_gate.py -v` → 3 passed
- [x] Negative-case verification: temporarily reverted the workflow → the lint test correctly fails with `.github/workflows/scheduled-tests.yml:99 partial marker (slow/long/performance) + --cov= without --cov-fail-under=0 / --no-cov-fail-under / --no-cov`. Restored the fix → all 3 pass.
- [x] Simulated the new slow-unit-tests step locally on `JuniperCascor1` (Python 3.13 + torch 2.11.0): `python -m pytest -m "unit and slow" src/tests/unit --slow --verbose --timeout=300 --maxfail=5 --junitxml=reports/junit/junit-slow-unit.xml` → `10 passed, 3621 deselected in 8.40s`, exit 0.
- [x] Pre-commit on changed files: all hooks pass (`SKIP=pytest-unit,coverage-gate pre-commit run --files ...`).
- [x] CI-equivalent unit suite locally with new test included: 0 failures, 0 errors.
- [ ] After merge: confirm next nightly Scheduled Long Tests run goes green.

### DIFF
--- a/.github/workflows/scheduled-tests.yml
+++ b/.github/workflows/scheduled-tests.yml
@@ -95,7 +95,17 @@ jobs:
           echo "║       Juniper Cascor - Slow Unit Tests                     ║"
           echo "╚════════════════════════════════════════════════════════════╝"
 
-          # Run unit tests marked as slow
+          # Run unit tests marked as slow.
+          #
+          # Intentionally no ``--cov=src``: only ~10 tests carry the ``slow``
+          # marker, covering ~34% of src/. pyproject.toml configures
+          # ``[tool.coverage.report] fail_under = 80`` which then causes
+          # coverage to exit 1 ("Required test coverage of 80.0% not reached.
+          # Total coverage: 33.75%") and fails the entire scheduled run even
+          # though all 10 tests pass. The full-suite coverage gate already
+          # runs in ci.yml's ``unit-tests`` job over the proper test set; the
+          # slow tests' value is correctness, not coverage. Enforced by
+          # ``src/tests/unit/test_workflow_coverage_gate.py``.
           python -m pytest \
             -m "unit and slow" \
             src/tests/unit \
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #309, opened 2026-05-26
fix(tests): isolate Settings from developer's local .env file

pcalnon:
## Summary

19 unit tests fail locally for developers who have populated `.env` files in the repo root (the gitignored copy of `.env.example` created by `cp .env.example .env`). CI never reproduces because runner checkouts have no `.env`. This PR fixes 18 of those 19 — the 19th (`test_version_matches_pyproject`) is a stale-editable-install issue resolved by `pip install -e ".[all]" --no-deps`, not a code bug.

## Root cause

`src/api/settings.py` configures pydantic-settings with `env_file=".env"`, so every fresh `Settings()` call reads the developer's local `.env`. pydantic-settings layers `.env` *under* `os.environ`, so a per-test `monkeypatch.delenv("JUNIPER_DATA_URL")` removes the OS-level value but leaves the value loaded from `.env` in effect.

Any test asserting "field default applies when no env var is set" then fails with a leaked value — concretely, `juniper_data_url='http://127.0.0.1:8100'` from a populated local `.env` when the test expected `None`.

## Affected tests (18 across 6 files)

| File | Failing tests |
|---|---|
| `src/tests/unit/test_cfg_04_juniper_data_url_settings.py` | 1 |
| `src/tests/unit/test_data_provider_coverage.py` | 3 |
| `src/tests/unit/test_spiral_data_provider.py` | 5 |
| `src/tests/unit/test_spiral_problem_juniper_data_integration.py` | 2 |
| `src/tests/unit/api/test_api_health.py` | 3 |
| `src/tests/unit/api/test_r2_1_4_wire_compat.py` | 4 |

All of class \"default-None / no-URL-configured / readiness-200-when-data-unconfigured\".

## Changes

- **`src/tests/conftest.py`** — new session-scoped autouse fixture `_disable_settings_env_file_for_tests` that sets `Settings.model_config[\"env_file\"] = None` for the test session and clears `get_settings`'s `lru_cache` before yielding. Restores the original `env_file` value on teardown. Tests continue to pick up values from `os.environ` (monkeypatched per test) and from explicit `Settings(...)` kwargs.
- **`src/tests/unit/test_env_file_isolation.py`** (new) — pinned regression tests:
  - `test_settings_env_file_is_none_during_test_session` — direct check that `Settings.model_config[\"env_file\"]` is `None` (proves the fixture ran).
  - `test_local_dot_env_in_cwd_does_not_leak_into_settings` — synthesizes a polluting `.env` in `tmp_path`, `monkeypatch.chdir`s into it, clears env vars, instantiates `Settings()`, and asserts class defaults apply. Survives refactors of the fixture mechanism — even if the implementation changes from `env_file=None` to a chdir+stub approach, this contract holds.

## Test plan

- [x] All 18 originally-failing tests pass when run from a worktree containing the user's `.env` (copied from `~/Development/python/Juniper/juniper-cascor/.env`)
- [x] New regression test passes after fix (2/2)
- [x] Negative-case verified: temporarily reverted the conftest fixture via `git stash`, both regression tests + 2 representative originally-failing tests fail with the exact bug signature (`Settings.model_config['env_file']` is `'.env'`; `juniper_data_url='http://leaked-from-dot-env:8100'`). Restored the fix, all pass.
- [x] CI-equivalent unit suite: `pytest -m \"unit and not slow\" src/tests/unit --timeout=60` — exit 0, 0 failed, 0 errors.
- [x] Pre-commit on changed files: all hooks pass.
- [ ] After merge: re-run user's `pytest --slow --integration --run-performance -vvv --cov .` from the main repo to confirm green.


pcalnon (commented):
approved

PR #311, opened 2026-05-27
fix: make api_keys validator accept None / empty / csv / list

pcalnon:
## Summary

- Replace `_empty_string_to_none` field-validator on `Settings.api_keys` with `_parse_api_keys`, mirroring the validator on juniper-data's `Settings`.
- The new validator accepts every reasonable input shape: `None`, empty string, csv string, list, and (via pydantic-settings auto-deserialise) JSON-array env var.
- Add 8 new regression tests in `src/tests/unit/api/test_api_settings.py` pinning each shape plus an end-to-end test that reads the bare `CHANGE_BEFORE_PRODUCTION_USE` placeholder from a tmp-path "secret file" exactly the way `get_secret` does in production.

## Root Cause

The pre-fix validator only handled the empty-string case:

```python
@field_validator("api_keys", mode="before")
@classmethod
def _empty_string_to_none(cls, v: Any) -> list[str] | None:
    if isinstance(v, str) and v.strip() == "":
        return None
    return v   # ← non-empty string falls through to pydantic's list[str] coercion
```

When the Docker secret file at `JUNIPER_CASCOR_API_KEYS_FILE` contained the bare deploy-side placeholder `CHANGE_BEFORE_PRODUCTION_USE`, the validator returned it unchanged and pydantic raised:

```
pydantic_core.ValidationError: 1 validation error for Settings
api_keys
  Input should be a valid list [type=list_type, input_value='CHANGE_BEFORE_PRODUCTION_USE', input_type=str]
```

→ `juniper-cascor` restart-loop on any fresh `docker compose --profile full up`.

## Contract Now (mirrors juniper-data)

| Input | Result |
|---|---|
| `None` | `None` (auth disabled) |
| `""` | `None` (empty secret file = auth disabled — documented state) |
| `"key1"` | `["key1"]` |
| `"key1,key2,key3"` | `["key1", "key2", "key3"]` |
| `"key1, key2 , ,"` | `["key1", "key2"]` (trim + drop empties) |
| `["key1", "key2"]` | `["key1", "key2"]` (pass-through; covers JSON env-var auto-parse + programmatic callers) |

## Test plan

- [x] `pytest src/tests/unit/api/test_api_settings.py -v` → **13 passed** (5 existing + 8 new).
- [x] `pytest src/tests/unit/api/` → **1361 passed, 0 failed** (whole api unit suite — no regressions).
- [ ] CI: full unit + integration suite + lint/format on the matrix.

## Related

- Companion juniper-deploy#91 wraps the placeholder in a JSON-array literal so older cascor images (without this validator) also work. After both ship, the placeholder shape no longer matters.
- Discovered after CVE-2026-48710 alertmanager port-conflict recovery (juniper-deploy host snap removed → cascor came up → exposed this contract mismatch). The end-to-end regression test in this PR is the direct fixture for that incident.

### DIFF
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -200,10 +200,32 @@ class Settings(BaseSettings):
 
     @field_validator("api_keys", mode="before")
     @classmethod
-    def _empty_string_to_none(cls, v: Any) -> list[str] | None:
-        if isinstance(v, str) and v.strip() == "":
+    def _parse_api_keys(cls, v: Any) -> list[str] | None:
+        """Normalise ``api_keys`` to ``list[str] | None``.
+
+        Accepts (in order of likelihood):
+
+        * ``None`` or empty string → ``None`` (auth disabled).
+        * Plain string ``"key1,key2"`` → ``["key1", "key2"]`` (the format
+          a single-string Docker secret file naturally produces; matches
+          the juniper-data parser).
+        * Already a list (including JSON-deserialised lists from
+          ``pydantic-settings``) → returned as-is.
+
+        Without this validator the bare string read from a Docker secrets
+        file (e.g. ``CHANGE_BEFORE_PRODUCTION_USE`` from
+        ``secrets.example/juniper_cascor_api_keys.txt``) would hit
+        pydantic's ``list[str]`` coercion and fail with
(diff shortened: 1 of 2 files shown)


pcalnon (commented):
approved

PR #315, opened 2026-05-30
fix(cascade_correlation): candidate result collection drops local-pool results -> 0 hidden units

pcalnon:
## Summary

On the deployed stack a training run trained **all** candidate units but grew **zero hidden units** and prematurely reported "Completed", with cascor logging `_process_training_results: Mismatch in results count: expected 40, got 2`. This fixes the two compounding root causes.

## Root cause (two compounding defects)

**Defect 1 — collection timeout shorter than candidate training time (affects local single-path *and* dual-path).** `_collect_training_results` used a fixed **60s total** deadline. A 40-candidate pool can take >2 min, so the collector abandoned the round with ~0 results while the persistent workers were still training → `best_candidate=None` → 0 hidden units. The live round logged `Training duration: 0:02:18`, and the ~60s gap between "Persistent pool created" and "Dispatching to remote" matched the deadline exactly.

**Defect 2 — dual-path remote-fallback recreated the local pool mid-round.** `TaskDistributor._execute_remote_with_fallback` retries failed/incomplete remote results via `local_fn` → `_execute_parallel_training(small_batch)` → `_ensure_worker_pool(2)`. The old exact-size check (`_persistent_pool_size == num_workers`) treated 15≠2 as invalid and tore the pool down — SIGKILLing the 15 in-flight workers and nulling their result queue.

## Fix

- **Defect 1:** `_collect_training_results` now waits on an **inactivity** deadline (reset on each result) + a **worker-liveness early-exit**, not a fixed total deadline. New constant `_CASCADE_CORRELATION_NETWORK_RESULT_COLLECTION_INACTIVITY_TIMEOUT = 300.0`; `workers` threaded through from `_collect_worker_results`.
- **Defect 2:** `_ensure_worker_pool` reuses a healthy pool whenever `num_workers <= _persistent_pool_size` (a larger pool serves a smaller request; surplus workers idle). It only rebuilds when no healthy pool exists or **more** workers are needed.

## Tests

- New `tests/unit/test_candidate_result_collection_dualpath_regression.py` — 7 tests pinning both defects (inactivity reset, liveness early-exit, alive-worker-keeps-waiting; pool reuse-smaller / reuse-equal / recreate-larger / recreate-degraded). No real worker processes spawned.
- Updated two `_collect_training_results` monkeypatch fakes for the new `workers=` kwarg.
- Green locally (JuniperCascor1): the new file + `test_parallel_training`, `test_task_distributor`, and `test_cascade_correlation_coverage{,_deep,_final}`.

[...]

PR #316, opened 2026-06-01
fix(monitor): track current_hidden_units from the metric stream (status stuck at 0)

pcalnon:
## Summary

Follow-up surfaced during the **#315 live-verify**: `GET /v1/training/status` reported `current_hidden_units: 0` for the *entire* run even as the cascade grew units (`/v1/network` showed the real count). **Canopy's status bar reads this field**, so a fully-working #315 looked broken from the dashboard.

## Root cause

`TrainingMonitor.current_hidden_units` is incremented **only** in `on_cascade_add` — and `on_cascade_add` has **no production caller** (a `grep` finds it invoked only in tests). The real per-epoch status updater, `on_epoch_end` (called from the manager's output-training callback with `hidden_units=len(network.hidden_units)`), recorded the count into the metric buffer but never updated `current_hidden_units`. So the field sat at its `0` init forever.

## Fix

`on_epoch_end` now sets `current_hidden_units` from that authoritative per-epoch count (the monitor is decoupled from the network, so the metric stream is the correct channel). The status tracks growth `0→1→2…` (updating after each unit-add's output-retrain pass), without depending on the unwired callback. `on_cascade_add` is left intact (still `+= 1`; its test still passes).

## Tests

- New `test_lifecycle_monitor.py::test_on_epoch_end_updates_current_hidden_units` (0 → set to 2 → set to 3 via the metric stream).
- Existing `test_on_cascade_add` unchanged; monitor + monitoring-hooks suites green (50 tests under `JuniperCascor1`).

## Noted follow-up (distinct, out of scope)

`on_cascade_add` being unwired in production also means the **`cascade_add` WS-broadcast events never fire** — WS clients get no per-unit-add notifications. Separate fix.

## Live-verify

The diagnosis was **confirmed live** on the running #315 image (status `current_hidden_units=0` while `/v1/network` reported 4). This fix is unit-tested; a full live-verify needs a cascor image rebuild — deferred to the next stack rebuild / post-merge.


pcalnon (commented):
approved

PR #320, opened 2026-06-03
feat(status): surface grow_network completion_reason in get_status (Issue #3 diagnosability)

pcalnon:
## Summary

**Producer side** of the Issue #3 diagnosability follow-up: cascor now records *why* a training run stopped growing and surfaces it in `/v1/training/status`, so canopy can distinguish a genuine **convergence** from a **0-unit stall** instead of both rendering a bare "Completed". The **canopy consumer is a separate follow-up** (as agreed — cascor first).

Item **B** of the canopy-diagnosability queue (A was already shipped in #341; C is canopy #353).

## What changed

- **`grow_network`** (`src/cascade_correlation/cascade_correlation.py`) sets `self._completion_reason` immediately before each existing loop exit — **assignment-only, no new branches and no control-flow change**, so the fragile cascade-termination logic is untouched (flake8 C901 unaffected). The `max_iterations` case is captured cleanly via a `for/else` clause.
- **`TrainingLifecycleManager.get_status()`** (`src/api/lifecycle/manager.py`) adds a top-level `"completion_reason"` key, read via `getattr(self.network, "_completion_reason", None)` so it's safe with no network loaded.
- The reason is also appended to the existing `grow_network: Finished training…` **INFO log**.

## The five reasons (1:1 with the loop exits)

| `completion_reason` | Trigger |
|---|---|
| `residual_collapsed` | residual error went `None` |
| `no_candidate` | no training results / no best candidate / no candidate met multi-select criteria — **the 0-unit stall signature** |
| `below_threshold` | best candidate below the adaptive correlation threshold |
| `early_stopped` | validation early-stop fired |
| `max_iterations` | growth loop hit its cap (`for/else`) |

Defaults to `None` before any training; resets at the start of each `grow_network` run.

## Testing

- **`src/tests/unit/test_completion_reason.py`** — drives `grow_network` to **each of the five exits** (via minimal mocks; `max_iterations` uses the training-free `max_iterations=0` path) and asserts the reason, plus the `None` default.
- **`test_lifecycle_manager.py::…::test_get_status_surfaces_completion_reason`** — `get_status()` surfaces the reason; `test_get_status` now also asserts the no-network `None` case.
- **132 cascade-correlation + lifecycle-manager tests green** (incl. the real-growth `grow_network` tests — no regression). Pre-commit clean (black/isort/flake8/mypy/bandit).

## Follow-up (separate PR, canopy)

canopy `_build_unified_status_bar_content` reads `completion_reason` from `/api/status` and renders e.g. *"Completed — converged"* vs *"Completed — stalled (0 new units)"* in the status detail.


pcalnon (commented):
approved

Issue #319, opened 2026-06-03
Distributed candidate training stalls every cascade growth: remote dispatch fails (38/2 split) → local fallback retry starves


pcalnon:
## Summary

On the deployed dual-path stack (local multiprocessing pool **+** remote WebSocket workers), cascade-correlation training **cannot grow past its first hidden unit**. Each cascade growth round splits the candidate pool 38-local / 2-remote, the **remote dispatch raises on every round** (caught and retried locally), and the **local fallback retry then starves** — `_collect_training_results` logs `Result queue empty, continuing` ~1×/s indefinitely while `best_correlation` keeps changing but `candidates_trained` freezes at `38/40` and the network never installs another unit.

This is the same _class_ of defect as cascor#315 (candidate result-collection) but on the **remote / dual-path leg**, which #315/#316 did not cover.

## Severity & impact

- **High** for any deployment with remote workers connected (`juniper-deploy` default = 2 workers). The network silently stops growing after 1 unit; training never converges. The monitor stays "training" forever.
- Single-process / no-remote-worker deployments are unaffected (all-local path).

## Environment

- juniper-cascor `main` @ `712772b` (includes #315 `f20a97c` + #316 `8ece31e`), image rebuilt 2026-06-02, version `0.5.0`.
- juniper-deploy full stack, 2 `juniper-cascor-worker` containers connected (dual-path), `candidate_pool_size=40`.

## Reproduction

```bash
KEY=$(cat juniper-deploy/secrets/juniper_cascor_api_keys.txt)
H=(-H "X-API-Key: $KEY" -H "Content-Type: application/json")
curl -s "${H[@]}" -X POST localhost:8201/v1/training/reset
curl -s "${H[@]}" -X POST localhost:8201/v1/training/start -d '{
  "epochs": 1000,
  "dataset": {"generator":"spiral","params":{"n_per_spiral":100,"n_spirals":2}},
  "params": {"max_hidden_units":3,"output_epochs":20,"candidate_epochs":30,"candidate_pool_size":40}
}'
# Network grows 0 -> 1, then stalls. Poll: status.monitor.current_hidden_units stays 1,
# candidates_trained frozen at 38/40, grow_iteration=0.
```

## Observed behavior (evidence)

Captured server logs from a live repro:
```
INFO  CascadeCorrelationNetwork: _execute_candidate_training: Training 40 candidates with 15 processes
[...]

PR #321, opened 2026-06-03
fix(distributed): map remote candidate results by global candidate_id (#319 dual-path stall)

pcalnon:
## Summary

Fixes the deployed-stack stall investigated in #319: with remote workers connected, cascade-correlation training **could not grow past one hidden unit**. Root cause is a local-list / global-index mismatch in the dual-path remote result conversion.

## Root cause

`TaskDistributor` splits a candidate pool larger than the local capacity, overflowing `N` tasks to remote workers (`N` = available workers; e.g. **2 of 40** → 38 local / 2 remote — this is the `candidates_trained 38/40` freeze). In `_dispatch_to_remote_workers`, `task_specs` holds only that **remote subset**, but each returned result was converted via:

```python
task_specs[tr.candidate_id]["candidate_data"]["input_size"]   # tr.candidate_id is the GLOBAL pool index
```

`tr.candidate_id` is the global pool index (0..`pool_size`-1, assigned at `_generate_candidate_tasks`, `cascade_correlation.py:2183`). For any remote candidate whose global id ≥ the subset length (the common case, e.g. ids 15/16), `task_specs[15]` on a 2-element list raises **`IndexError`**. That exception was swallowed by `TaskDistributor._execute_remote_with_fallback` as a bare `str(e)` (the masked `"(1)"`) and forced a **local-retry fallback every round, which then starved** (`_collect_training_results` looping `Result queue empty, continuing`), pinning the network at one unit.

Confirmed live: `Remote execution failed (1) — retrying 2 tasks locally`, 59× `Result queue empty`, `hidden_units=1`.

## Fix

1. **`cascade_correlation.py`** — build a `candidate_index → task_spec` map and look up each result by identity; defensively **skip** results whose `candidate_id` isn't part of this dispatch (also guards the round_id-less stale-result contamination gap noted in #319).
2. **`task_distributor.py`** — log `repr(e)` + `traceback` in the remote-fallback handler so this class of failure is no longer masked as `"(1)"`.

## Testing

- New `TestRemoteDispatchCandidateIdMapping` in `test_candidate_result_collection_dualpath_regression.py` drives the **real** `_dispatch_to_remote_workers` with global candidate ids (15,16 of a 40-pool):
  - **Proven regression guard** — both new tests fail with `IndexError: list index out of range` on the pre-fix source (verified via `git stash` of just the fix), pass after.
  - Plus an unknown-candidate-id skip test for the stale-result hardening.
- `test_candidate_result_collection_dualpath_regression.py` (incl. the #315 tests): **9 passed**.
- `test_task_distributor.py` + `test_parallel_training.py`: green.
- pre-commit (black/isort/flake8/mypy/bandit): clean.
- **Live verification on the deployed dual-path stack: in progress** (rebuilding cascor from this branch; will confirm the network grows past 1 unit and update this PR).

## Scope / follow-ups (tracked in #319)

This resolves the **primary** root cause (remote-dispatch crash → fallback starvation). Residual hardening remains for #319:
- the local-retry fallback should not be able to starve even when remote *legitimately* fails (the `_collect_training_results` path);
- add `round_id` to the remote `TaskResult` to fully close the RC-5 stale-round gap on the remote leg.

## Requirements

References #319. No tracked JR-ID directly applies.

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -1166,7 +1166,14 @@ class CascadeCorrelationNetwork:
 
         # Convert internal task tuples to wire protocol task specs
         task_specs = []
-        for _task_idx, candidate_data_tuple, training_inputs in tasks:
+        # ISSUE-319 FIX (the actual stall trigger): source the per-round candidate
+        # training params from the network config, NOT from positional indexing into
+        # ``training_inputs``. Under the OPT-5 SharedMemory path ``training_inputs`` is a
+        # *dict* (shm metadata), so ``training_inputs[1]`` raised ``KeyError(1)`` (masked
+        # as the bare "(1)") on every dual-path round, forcing the remote dispatch into a
+        # starving local-retry fallback. These are per-round constants — the OPT-5 dict
+        # itself is populated from these same ``self.candidate_*`` attributes.
+        for _task_idx, candidate_data_tuple, _training_inputs in tasks:
             candidate_index, input_size, activation_name, random_value_scale, candidate_uuid, candidate_seed, random_max_value, sequence_max_value = candidate_data_tuple
             task_specs.append(
                 {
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #322, opened 2026-06-03
fix(distributed): dual-path remote collection budget + idle-worker dispatch — #319 defects 3-5 (cascor side)

pcalnon:
## Summary

Fixes three cascor-side defects in the **dual-path** (local pool + remote WebSocket workers) candidate-training leg that prevented remote workers from contributing to cascade growth. Companion to #321 (which fixed the two remote-dispatch crashes). All three are verified at the cascor↔worker boundary on the live deployed stack.

> ⚠️ **Partially addresses #319.** These complete the **cascor** side. End-to-end remote candidate *execution* is still blocked by a separate **worker-side** defect — see [Scope / CW-05](#scope--cw-05). Filed: pcalnon/juniper-cascor-worker#97.

### #3 — remote result-collection budget (wrong timeout)
`_dispatch_to_remote_workers` waited on `collect_results(timeout=candidate_training_shutdown_timeout)` — **~10s**, a *process-teardown* budget — for a full candidate round that takes tens of seconds. The wait always expired, every remote result was dropped as "late", and the 2 remote tasks were retried on the saturated 15-process local pool.

- New `CascadeCorrelationNetwork._remote_result_collection_timeout()` scales the wait to `candidate_epochs` (1s/epoch) with a **120s floor / 900s ceiling**.
- `WorkerCoordinator.collect_results` is now a poll loop with a **worker-liveness early-exit**: returns promptly once no workers remain connected, so the larger budget can't hang a round. It still returns the instant all results arrive — healthy rounds are unaffected.

### #4 — remote result round isolation
`TaskResult` carries a `round_id`, attached server-side in `submit_result` from the dispatching `PendingTask`. `_dispatch_to_remote_workers` filters by it, so a late result for a still-tracked stale `_pending_tasks` entry can't leak into a later round (the remote analogue of #315's RC-5).

### #5 — idle workers never got tasks submitted after they connected (the unlock)
`_try_dispatch_task` ran **only** at worker-connect and after a task result, so an idle worker sitting in the WS message loop had no trigger to pull tasks submitted mid-session — they sat unassigned until the budget expired → local-retry fallback every round. The **heartbeat handler now dispatches pending work to idle workers**, guarded on registry idle state (`get_next_assignment` doesn't itself refuse a busy worker). Delivery latency ≈ one heartbeat interval.

## Live verification (deployed dual-path stack, 2 workers)

| Claim | Evidence |
|---|---|
| #3 budget applied | cascor: `Remote returned 0/2 results (120.0s)` — waited the new budget, not 10s |
| #5 delivery works | cascor `Dispatching 2 tasks` @ `21:34:13.093` → worker `Received task … (3 tensors)` @ `21:34:13.908` (**0.8s**). Before the fix: workers received nothing (`tasks_completed=0`, zero execution attempts) |

## Scope / CW-05

Live-verify surfaced the *true* end-to-end blocker, which is **not** in cascor: the `juniper-cascor-worker` container **cannot import the cascor `candidate_unit` codebase** (launched with no `--cascor-path`, no cascor src mount, no cascor package), so it crashes on every task it's handed:

```
ModuleNotFoundError: No module named 'candidate_unit'   # juniper_cascor_worker/task_executor.py:46
```

So the remote tier has effectively never *executed* a candidate in this deployment; all observed "dual-path" growth (incl. #315's 0→2) was the local-retry path. With CW-05 open, these cascor fixes are correct but cannot yet make the remote tier contribute — and #3 means each round waits 120s (vs 10s) before local fallback while workers crash-loop. **Tracked: pcalnon/juniper-cascor-worker#97** (stopgap: deploy mounts cascor src + `--cascor-path`; canonical: `juniper-cascor-core` package).

## Testing

- `pytest test_worker_stream.py test_worker_coordinator.py test_candidate_result_collection_dualpath_regression.py` → **68 passed**.
  - `TestHeartbeatDispatch`: task delivered to an idle worker via heartbeat (the #319 scenario); busy-worker guard.
  - `TestCollectResults`: `round_id` attachment; liveness early-exit returns promptly when workers vanish.
  - dual-path regression: timeout scales to training (not shutdown); stale-round result discarded.
- `pre-commit run` on all changed files: black, isort, flake8, **async-route audit**, mypy, bandit — all pass.

## Commits

1. `#3` collection budget + liveness early-exit, `#4` round_id isolation.
2. `#5` heartbeat-triggered idle-worker dispatch + CHANGELOG.

Refs #319 · Refs pcalnon/juniper-cascor-worker#97


pcalnon (commented):
approved

PR #324, opened 2026-06-07
fix(distributed): int-coerce remote candidate bounds; backport cascor-core normalization (#319)

pcalnon:
## Summary

Completes the **#319 dual-path remote candidate** fix and restores the
**juniper-cascor-core `package == src` drift invariant** — without touching the
(separately, actively-published) `juniper-cascor-core` package.

## Root cause — remote candidate crash

`_dispatch_to_remote_workers` serialized the **int-valued** candidate bounds
`random_max_value` / `sequence_max_value` with `float()`, so the JSON wire carried
`1.0` instead of `1`. The remote worker fed those floats to `random.randint()` /
`range()` → `"'float' object cannot be interpreted as an integer"`, starving the
remote leg into the local-retry fallback (the dual-path stall surfaced during
CW-05 stopgap verification). Fix: send them as `int()`.

## Backport — drift invariant

The worker-payload normalization that `juniper-ml#370-376` added **directly to the
`juniper-cascor-core` package** (`CandidateUnit._coerce_int_like` +
`ActivationWithDerivative._normalize_activation_fn`) is backported **verbatim into
cascor src**, restoring the `package == src` invariant that the drift-guard
(`juniper-ml tests/test_cascor_core_drift.py`) enforces. cascor src is the single
source of truth the package extracts from, so the behavior now lives where it
belongs and is tested at the source. The dispatch `int()` fix makes the model-side
coercion **belt-and-suspenders** rather than load-bearing.

> **Sync point:** `candidate_unit.py` / `activation.py` here are byte-identical to
> the `juniper-cascor-core` package at `juniper-ml@f4d66db` (#384). If the package's
> copies of those two files change again before this merges, re-sync is needed.

## Changes

- `src/cascade_correlation/cascade_correlation.py` — `float()`→`int()` for the two
  remote candidate bounds in the dispatch payload.
- `src/candidate_unit/candidate_unit.py`, `src/utils/activation.py` — backport the
  worker-facing normalization (byte-identical to the package).
- `src/tests/unit/test_candidate_result_collection_dualpath_regression.py` — guard
  that dispatched bounds stay `int` on the wire.
- `src/tests/unit/test_candidate_core_worker_contracts.py` *(new)* — source-side
  contracts for the worker-facing normalization (tuple activation, float-valued int
  bounds, bool/non-int rejection), mirroring the package smoke tests.

## Testing

- Full cascor unit suite: **3689 passed, 0 failed**, 15 skipped (`--slow`).
- Serialization/pickle (`--integration --slow`): **62 passed** — the backport touches
  `__getstate__`/`__setstate__`, and cascor's local pool pickles candidates.
- New contracts + dual-path regression: **18 passed**.
- juniper-ml drift-guard re-run against this branch: **passes** (invariant restored).
- pre-commit (black / isort / flake8 / mypy / bandit): clean.

### DIFF
--- a/src/candidate_unit/candidate_unit.py
+++ b/src/candidate_unit/candidate_unit.py
@@ -44,6 +44,7 @@ import random
 import sys
 import uuid
 from dataclasses import dataclass, field
+from numbers import Integral, Real
 
 # from typing import Optional
 from typing import Any, Callable, Optional
@@ -141,6 +142,16 @@ from utils.activation import ActivationWithDerivative  # noqa: F401, E402
 
 #####################################################################################################################################################################################################
 class CandidateUnit:
+    @staticmethod
+    def _coerce_int_like(value: Any, field_name: str) -> int:
+        """Normalize JSON-decoded numeric payload fields used by seed/range APIs."""
+        if isinstance(value, bool):
+            raise TypeError(f"{field_name} must be an integer, not bool")
+        if isinstance(value, Integral):
+            return int(value)
+        if isinstance(value, Real) and float(value).is_integer():
+            return int(value)
+        raise ValueError(f"{field_name} must be an integer-valued number, got {value!r}")
 
     #################################################################################################################################################################################################
     # Initialize a candidate unit.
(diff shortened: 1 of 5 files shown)


pcalnon (commented):
approved

PR #242, opened 2026-05-10
feat(dataset): stage/cancel pending dataset + reload on start (Issue #3 Phase 1, PR-6)

pcalnon:
## Summary

Server-side half of [FRONTEND_ISSUES_PLAN_2026-05-09 §3.5.1 + §3.5.2 Phase 1](https://github.com/pcalnon/juniper-canopy/blob/main/notes/FRONTEND_ISSUES_PLAN_2026-05-09.md). Closes Issue #3's "Dataset View tab doesn't affect training" by giving the dashboard's Apply Dataset button a real backend.

> **Cross-repo lockstep partner: PR-7 (canopy UI + adapter wiring).** Independently safe to merge first since no canopy code yet calls the new endpoints — they're wired but unused on main until PR-7 lands.

### Three new REST routes (`api/routes/training.py`)

| Method | Path | Purpose |
|--------|------|---------|
| `POST` | `/v1/training/dataset` | Stage a config; returns `{status: "staged", config: {...}}`. Empty body clears (idempotent with DELETE). |
| `DELETE` | `/v1/training/dataset` | Phase 1 Cancel button target; returns `{status: "cleared", discarded: …}` so the canopy toast can show what was reverted. |
| `GET` | `/v1/training/dataset/pending` | Peek for the canopy banner; `{pending: null \| {...}}`. |

`StageDatasetRequest` body is fully optional (`dataset_type` Literal[spirals, xor, mnist, circles, moons], `n_samples`, `noise`, `rotations`, `n_spirals`) so canopy can build the body from whatever subset of dataset inputs the user touched.

### Lifecycle manager (`api/lifecycle/manager.py`)

- New `_pending_dataset_config: Optional[Dict[str, Any]]` initialised to `None`.
- New thread-safe `stage_dataset_config(**cfg)` / `clear_pending_dataset_config()` / `get_pending_dataset_config()` accessors.
- New `_reload_dataset(**cfg)` private method — instantiates a `JuniperDataClient` from env vars (mirrors `api/app.py::_auto_start_training`'s pattern), `create_dataset(generator=cfg['dataset_type'], params=…)`, `download_artifact_npz`, swaps `_train_x/_train_y` (and val tensors if the artifact carries `X_test`/`y_test`).
- `start_training` consumes the staged config before submitting the training future. **On reload failure the staged config survives** so the user can fix the upstream juniper-data issue and Restart-and-retry without losing their selection.
[...]

PR #331, opened 2026-06-14
fix(api): resolve juniper-data API key on live dataset swap (502/401 root cause)

pcalnon:
## Summary

Fixes the root cause of `POST /v1/training/dataset/live` → **`502 juniper-data fetch failed: Request failed (401): Missing API key`**. cascor never sent its juniper-data API key on the dataset-reload path.

`_reload_dataset` resolved the key via:
```python
try:
    from secrets_util import get_secret   # ← module does NOT exist in this repo
except ImportError:
    get_secret = lambda _key: None        # ← always taken → api_key = None
```
So `JuniperDataClient` was built with `api_key=None`, sent no `X-API-Key`, and juniper-data answered 401 → cascor 502. **Both linters that would have caught it were suppressed inline** (`# type: ignore[import-not-found]` + `# noqa: E731`). Introduced in #242 (`30b2f0f`).

The same key resolves correctly everywhere else — inbound auth (`settings.py:196`) and the auto-start fetch (`app.py:316`) both use `from api.secrets import get_secret`, which honors `JUNIPER_DATA_API_KEY_FILE`. **The fix routes the live-swap path through that same correct helper.**

## Changes

- **`src/api/lifecycle/manager.py`** — `_reload_dataset`: `from api.secrets import get_secret` (drops the nonexistent-module import, the silent `None`-lambda degradation, and the two linter suppressions).
- **`src/tests/integration/api/test_pending_dataset.py`** — `test_reload_dataset_resolves_juniper_data_api_key_from_secret_file`: asserts the resolved `JUNIPER_DATA_API_KEY_FILE` value reaches `JuniperDataClient(api_key=…)` (fails on the old `None`-key code). All 13 in the file pass.
- **`notes/CASCOR_STARTUP_SECRET_INDIRECTION_INVESTIGATION_2026-06-14.md`** — full investigation: this bug **plus** the related startup/FSM/API/Canopy defects surfaced during the #319 live demo (auto-start-on-boot default-ON, FSM 409-after-stop, experimental-gate default-OFF, juniper-data-client `_FILE`-blindness, opaque Canopy training-control errors), each root-caused with file:line and triaged for follow-up.

## How it surfaced

Verifying CW-05 dual-path (#319) end-to-end, attempting to load a fresh dataset hit four gates; this 401/502 was the last. The dual-path itself is already proven (2/2 remote tasks). See the notes doc §2.

## Validation

- `pytest src/tests/integration/api/test_pending_dataset.py --integration` → **13 passed**
- pre-commit (black, isort, flake8, **mypy**, bandit, doc-links) → all pass (the fix *removes* the prior linter suppressions, so mypy/flake8 now see the import for real)
- `grep -r secrets_util src` → only explanatory comments remain; no import

## Scope / follow-up

Deliberately tight: the import-class fix is definitive and low-risk. The behaviour-changing and cross-repo items (auto-start default flip, FSM-409 messaging, juniper-data-client `_FILE` awareness, Canopy error surfacing, an import-resolution CI guard) are triaged in the notes doc §5 for separate PRs so each is reviewed on its own blast radius.

### DIFF
--- a/notes/CASCOR_STARTUP_SECRET_INDIRECTION_INVESTIGATION_2026-06-14.md
+++ b/notes/CASCOR_STARTUP_SECRET_INDIRECTION_INVESTIGATION_2026-06-14.md
@@ -1,0 +1,263 @@
+# CasCor Startup / Secret-Indirection / FSM / API / Canopy — Investigation & Fix
+
+**Project**: Juniper — juniper-cascor
+**Author**: Paul Calnon (investigation driven via Claude Code)
+**Date**: 2026-06-14
+**Status**: Root-caused. Critical fix implemented in this branch; related issues triaged below.
+**Branch**: `fix/cascor-data-api-key-file-indirection`
+
+---
+
+## 1. Executive summary
+
+While verifying the CW-05 dual-path remote-worker fix (juniper-cascor#319) end-to-end on the live
+deploy stack, an attempt to load a fresh dataset and grow the network surfaced a cluster of
+**service-config / startup-behaviour defects** in juniper-cascor. They are individually small but
+collectively a *major blocker* for any workflow that drives cascor through its API or through the
+Canopy dashboard (iterative training, live dataset swap, automated/CI runs).
+
+The central defect — and the one this branch fixes — is a **silent secret-indirection failure**:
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #332, opened 2026-06-14
fix(api): auto-start off by default + specific start-failure reason (startup hardening)

pcalnon:
## Summary

Two low-risk cascor startup/lifecycle hardening changes — follow-up #1 to the secret-indirection investigation (#331). Both address the "faulty startup behaviour" surfaced during the #319 dual-path live demo.

### 1. `auto_start` default → `False`

`settings.py` defaulted `auto_start = True`. On boot the lifespan handler trains onto a **default (empty) network**, so every fresh cascor comes up already-`STARTED` on a 0-unit network — violating the clean-`STOPPED` initial-state assumption that API / Canopy / automation callers make. (This is exactly why the #319 probe found `epoch 7576 / hidden_units 0` on a "fresh" stack.) The code itself warns it is *"demo/dev only"* yet defaults it on.

`juniper-deploy/docker-compose.yml` sets `JUNIPER_CASCOR_AUTO_START: "true"` **explicitly**, so flipping the code default does **not** change the demo stack — it only makes a bare cascor boot clean.

### 2. `POST /v1/training/start` 409 names the actual reason

The route flattened *every* start failure — no network, training already active, **no dataset loaded**, **juniper-data fetch failed**, investigating/replaying — into one generic `"Training cannot be started in the current state"`. That masked the secret-indirection 401 (#331) as a bogus *state* error (the "FSM-409-after-stop" from the demo was that fetch failing). Now the 409 detail carries the specific cause, so API + Canopy callers can act on it.

## Changes
- `src/api/settings.py` — `_JUNIPER_CASCOR_API_AUTO_START_DEFAULT` → disabled.
- `src/api/routes/training.py` — `start_training` 409 detail → `f"Training cannot be started: {e}"`.
- tests — `test_auto_start_default_is_off` (fresh `Settings().auto_start is False`); `test_start_training_409_surfaces_specific_reason` (the 409 names "Training data not provided"). 45/45 in the two touched test files pass.

## Validation
- targeted tests: **45 passed**
- pre-commit (black, isort, flake8, **mypy**, bandit) — all pass
- backward-compatible: existing `test_start_training_no_body` (`"cannot be started"`) still passes

## Context
Part of the startup-defect cluster triaged in `notes/CASCOR_STARTUP_SECRET_INDIRECTION_INVESTIGATION_2026-06-14.md` (§3.3, §3.4) — that doc lands in #331. Next in the series: import-resolution CI guard, then cross-repo (juniper-data-client `_FILE`, Canopy error surfacing).

### DIFF
--- a/src/api/routes/training.py
+++ b/src/api/routes/training.py
@@ -74,8 +74,14 @@ async def start_training(request: Request, body: TrainingStartRequest = None) ->
         result = lifecycle.start_training(x=x, y=y, x_val=x_val, y_val=y_val, **kwargs)
         return success_response(result)
     except (RuntimeError, ValueError) as e:
+        # Surface the specific reason (no network created / training already in progress /
+        # no dataset staged ("Training data not provided") / juniper-data fetch failed /
+        # investigating|replaying a snapshot) rather than a generic message, so API and
+        # Canopy callers can tell *why* the start was rejected and act on it. The generic
+        # string previously masked a juniper-data fetch failure as a bogus "state" error.
+        # See notes/CASCOR_STARTUP_SECRET_INDIRECTION_INVESTIGATION_2026-06-14.md (3.4).
         logger.debug("Start training failed: %s", e)
-        raise HTTPException(status_code=409, detail="Training cannot be started in the current state") from e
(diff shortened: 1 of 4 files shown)


pcalnon (commented):
approved

PR #333, opened 2026-06-14
feat(provenance): git_sha/build_date on /v1/health + OCI labels (obs 0.4.0)

pcalnon:
## Summary

Adds **build provenance** to juniper-cascor — the running service now reports the source `git_sha` and ISO-8601 `build_date` baked into its image at build time, on `/v1/health` and `/v1/health/ready`, in the Prometheus `juniper_cascor_build` Info metric, and as OCI image labels. This is the foundation for ecosystem stale-image-drift detection (`make doctor`), addressing the recurring "running container is silently behind source" failure mode.

Mirrors the sibling **juniper-data PR #180** (the reference implementation) adapted to cascor, and the Dockerfile/accessor template from **juniper-cascor-worker #103**. Design of record: juniper-ml [`notes/BUILD_PROVENANCE_DESIGN_2026-06-14.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/BUILD_PROVENANCE_DESIGN_2026-06-14.md) §5.2 (Maximal scope, ratified 2026-06-14), tracked as juniper-ml#412.

## Changes

- **Dockerfile** (runtime stage): `ARG GIT_SHA`/`BUILD_DATE`/`APP_VERSION`; OCI labels `org.opencontainers.image.revision`/`.created`/`.version`; `ENV JUNIPER_CASCOR_GIT_SHA`/`_BUILD_DATE`. ARGs re-declared in the runtime stage (they don't cross stages).
- **`src/api/provenance.py`** (new): `git_sha()` / `build_date()` accessors reading the env vars — both `None` outside a provenance-stamped image (local dev / bare `docker build`); never raise. Mirrors `juniper_data/provenance.py`.
- **`/v1/health`** (`src/api/routes/health.py`): the hand-built dict now carries `git_sha` + `build_date`.
- **`/v1/health/ready`**: the shared `ReadinessResponse` now carries `git_sha` + `build_date` (obs 0.4.0 added these optional fields).
- **`set_build_info(...)`** (`src/api/app.py`): now passes `git_sha=` / `build_date=` (Prometheus `juniper_cascor_build` Info metric).
- **OQ-1**: `_API_VERSION` in `health.py` migrated from the hardcoded `"0.5.0"` literal to `importlib.metadata.version("juniper-cascor")` with `PackageNotFoundError` fallback, so the health-route version can no longer drift from `pyproject.toml`. (`app.py`'s `_API_VERSION` already used this pattern.)
[...]

PR #334, opened 2026-06-15
test(api): adapt route-introspection tests to fastapi 0.137 _IncludedRouter

pcalnon:
## Summary

fastapi **0.137.0** (released 2026-06-14) changed `app.include_router(...)` to emit `_IncludedRouter` route objects with no `.path`, so route paths are no longer flat in `app.routes`. This broke two tests in `TestAppFactoryConfigurations`:
- `test_rest_routes_registered` — asserted on `r.path`
- `test_websocket_routes_registered` — `AttributeError: '_IncludedRouter' object has no attribute 'path'`

**juniper-cascor `main` is red** on these (CI floats deps via the `.[all]` install, so it picks up 0.137.0 regardless of the lockfile), and every open PR inherits it. The app routes correctly; this is **test-only**.

## Fix

- `test_rest_routes_registered` → assert against `app.openapi()["paths"]`.
- `test_websocket_routes_registered` → new module-level `_collect_route_paths()` helper that descends into `_IncludedRouter` via `.include_context.included_router.routes` (WebSocket routes are absent from the OpenAPI schema). The helper handles both old (`.routes`) and new (`_IncludedRouter`) fastapi layouts.

## Testing

`pytest src/tests/unit/api/test_api_app_coverage_deep.py` → **26 passed** under fastapi 0.137.0 (reproduced the failure first, then confirmed green). black + isort + test-lane flake8 clean.

## Context

Kept separate from the build-provenance PR (#333) per route-test-vs-feature separation. Once this lands on `main`, #333 goes green via merge-with-main CI.

### DIFF
--- a/src/tests/unit/api/test_api_app_coverage_deep.py
+++ b/src/tests/unit/api/test_api_app_coverage_deep.py
@@ -24,6 +24,30 @@ from api.settings import Settings
 pytestmark = pytest.mark.unit
 
 
+def _collect_route_paths(routes, prefix=""):
+    """Collect all route paths (REST + WebSocket), descending into routers.
+
+    fastapi >=0.137 wraps ``app.include_router(...)`` results in
+    ``_IncludedRouter`` objects that have no ``.path``; their sub-routes live
+    under ``.include_context.included_router.routes``. WebSocket routes are also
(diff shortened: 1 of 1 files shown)


pcalnon (commented):
approved

PR #339, opened 2026-06-15
build(docker): add .dockerignore excluding nested egg-info (provenance hardening)

pcalnon:
## Summary

cascor had **no `.dockerignore`** — so the full build context was sent and any build artifact under `src/` would be `COPY src/`'d into the image. Because cascor runs from `/app/src` (`ENV PYTHONPATH=/app/src`), a stale `src/*.egg-info` would land on the import path ahead of site-packages and shadow `importlib.metadata.version("juniper-cascor")` — the exact class of bug fixed for juniper-canopy in [canopy #362](https://github.com/pcalnon/juniper-canopy/pull/362) (surfaced by the build-provenance `make doctor` work).

**cascor is not vulnerable today** — verified: its egg-info sits at the repo root (which `COPY src/` does not pick up), the running image has no egg-info on `/app/src`, and `/v1/health` reports the correct 0.5.0. So this is **preventive hardening** + general build-context hygiene.

## Change

New `.dockerignore` excluding build artifacts (`**/*.egg-info/`, `**/*.dist-info/`, `__pycache__/`, `*.py[cod]`, `build/`, `dist/`), VCS metadata (`.git/`), and test/type-check caches. The nested `**/`-prefixed egg-info/dist-info forms are the key bit — a root-only `*.egg-info/` (what canopy carried) does **not** match `src/*.egg-info`.

## Verification

- Pattern proven against a synthetic context containing `src/X.egg-info` — excluded at build time, while real source is kept.
- `src/tests/unit/test_dockerignore_egg_info.py` (3 tests) pins the nested-exclusion patterns so the hardening can't silently regress.
- CI's **Docker Build & Smoke Test** lane rebuilds + smoke-tests the image with the new build context.

## Notes

- Intentionally does **not** exclude `src/tests/` — kept minimal/safe (image behavior unchanged beyond artifact removal); a fuller hygiene `.dockerignore` is a separate optional follow-up.
- Build-provenance follow-up; the companion preventive change for juniper-data is a separate PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+# =============================================================================
+# JuniperCascor — Docker build context exclusions
+# =============================================================================
+# Keep build artifacts and VCS metadata out of the image. The nested ``**/``
+# egg-info / dist-info forms matter specifically because the Dockerfile does
+# ``COPY src/ ./src/`` and runs with ``ENV PYTHONPATH=/app/src`` — a stale
+# build artifact under src/ would otherwise land on the import path ahead of
+# site-packages and shadow the installed package's metadata version. That is
+# the class of bug fixed for juniper-canopy in #362 (see juniper-ml
+# notes/BUILD_PROVENANCE_DESIGN_2026-06-14.md). A root-only ``*.egg-info/`` is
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #390, opened 2026-07-06
fix(ci): docker smoke — opt into the published-container bind contract (heals main)

pcalnon:
## Summary

- Heals main's red **Docker Build & Smoke Test** (failing since the #379 merge, e.g. run 28757261573): #379 changed the image's baked default to `JUNIPER_CASCOR_HOST=127.0.0.1` (safe bare-container default under the SEC-F22 bind guard) but CI's smoke `docker run` was never opted into the new published-container contract. The container bound loopback internally — its in-container healthcheck passed, but the runner-side **Verify Health Endpoint** curl through the published port (`-p 127.0.0.1:8200:8200`) could not connect.
- Fix: add `-e JUNIPER_CASCOR_HOST=0.0.0.0` to the smoke's `docker run` (attestation was already present) — exactly the invocation #379's own README example documents for published containers. Publishing stays on the runner's loopback; no security-posture change.
- One file, +6 lines (env line + explanatory comment).

## Test plan

- [x] Root cause verified from run 28757261573 (failed step: `Verify Health Endpoint`; pre-#379 main run 28756532664 was green, both post-#379 runs red on exactly this job)
- [ ] This PR's own **Docker Build & Smoke Test** check passes (the fix proves itself in CI)
- [ ] All other checks green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LvPNTfp4ydj1hPSSFudKoM

PR #391, opened 2026-07-06
fix(cascor): repair latent snapshot restore/listing defects, dead accuracy guard, unexercised logging init

pcalnon:
## Summary

Repairs four verified latent defects in `cascade_correlation.py` / `snapshot_utils.py` — dead snapshot restore, dead snapshot listing, a dead `None` guard, and a suite-wide logging-init bypass — and pins the real success behavior with regression tests that demonstrably fail against the pre-fix code. Defects were independently verified on `main` @ `0a44938` (this branch's base).

### 1. `restore_snapshot` could never succeed (dead success path)

`restore_snapshot` is a `@classmethod`; its success path executed `cls.__dict__.update(loaded_network.__dict__)`. A class's `__dict__` is a read-only `mappingproxy` with no `.update`, so every load-then-restore raised `AttributeError`, landed in the `except`, and returned `False`. `cls._load_from_hdf5(...)` already returns a usable network before the broken copy step.

**Fix**: return the loaded network on success (mirrors the sibling classmethod `load_from_hdf5`, which returns network-or-`None`). All existing call sites invoke the method on the class (`CascadeCorrelationNetwork.restore_snapshot(...)`), so an instance-method conversion would have broken 7 existing test call sites. All failure paths keep the pinned `False` contract; return annotation and docstring updated to `Union["CascadeCorrelationNetwork", bool]`.

### 2. `list_hdf5_snapshots` always returned `[]`

Line 5014 called `HDF5Utils.list_hdf5_files(directory)`, which was **not defined anywhere in the codebase**, so every existing-directory call raised `AttributeError` and fell through the `except` to `return []`. (The nearest capability, `list_networks_in_directory`, returns verification dicts — a different contract than the documented "List of HDF5 file paths".)

**Fix**: implement `HDF5Utils.list_hdf5_files(directory) -> List[Path]` in `src/snapshots/snapshot_utils.py` matching the class's static-method idiom: sorted `.h5`/`.hdf5` paths, empty list for a missing directory. `list_hdf5_snapshots` itself is unchanged. Also updated the now-stale "finding" note in `test_serialization_error_paths_coverage.py`'s module docstring, which had documented this exact defect.

### 3. Dead `None` guard in `calculate_accuracy`

Lines 5375–5376 default `x`/`y` via the tuple-index idiom, so the `if x is None or y is None:` guard below could never fire. Removed the dead block; the code now has one real defaulting path plus the live isinstance/shape validation. Behavior for valid tensors is unchanged, and the existing `test_accuracy_none_inputs` semantics (None → `ValueError`) are preserved because that error comes from the live shape-mismatch branch, not the dead guard.

### 4. Real `_init_logging_system` was never executed by the unit suite

The autouse session fixture `_cache_logging_system` monkey-patches `_init_logging_system` away for the entire unit run, so regressions in the real body were invisible.

**Fix**: the session fixture now stashes the genuine method in a module global, exposed via a new function-scoped `real_init_logging_system` fixture; a dedicated test invokes the real body against a `tmp_path` log config. The suite-wide fast-logging speedup is fully preserved (the autouse patching behavior is untouched). Coverage proof below.

## Regression tests (`src/tests/unit/test_latent_defect_repairs.py`)

- **Round-trip restore**: saves a real snapshot to `tmp_path` (no serializer mocking), restores it, and asserts restored state matches the original (`input_size`, `output_size`, hidden-unit count, `torch.equal` on `output_weights`/`output_bias`) — not merely `is not None`.
- **Listing**: creates `.h5`/`.hdf5`/`.txt` files in `tmp_path`, asserts the returned list is exactly the two HDF5 paths (network-level and `HDF5Utils`-level), plus the preserved empty-list contracts.
- **Accuracy**: valid-tensor behavior and the shape-mismatch `ValueError` pinned.
- **Real logging init**: every assertion distinguishes the genuine body from the fast fixture (fresh `LogConfig` built from *this* config's `tmp_path`, `logger is log_config.get_logger()`, not a `_NoOpLogger`).

### Fail-first evidence (new tests vs. pre-fix code)

Run with only the tests + conftest fixture in place, source unfixed:

```text
FAILED tests/unit/test_latent_defect_repairs.py::TestRestoreSnapshotRoundTrip::test_restore_snapshot_round_trip_returns_restored_network
  AssertionError: restore_snapshot returned False for a valid snapshot (the pre-fix dead success path)
FAILED tests/unit/test_latent_defect_repairs.py::TestListHdf5Snapshots::test_list_hdf5_files_returns_hdf5_paths
  AttributeError: type object 'HDF5Utils' has no attribute 'list_hdf5_files'
FAILED tests/unit/test_latent_defect_repairs.py::TestListHdf5Snapshots::test_list_hdf5_files_missing_directory_returns_empty
  AttributeError: type object 'HDF5Utils' has no attribute 'list_hdf5_files'
FAILED tests/unit/test_latent_defect_repairs.py::TestListHdf5Snapshots::test_list_hdf5_snapshots_returns_files_in_existing_directory
  AssertionError: list_hdf5_snapshots returned a wrong/empty list (the pre-fix except fallback)
==================== 4 failed, 5 passed, 1 warning in 1.66s ====================
```

## Acceptance evidence (fixed code, env `JuniperCascor1`)

New tests:

```text
tests/unit/test_latent_defect_repairs.py .........                       [100%]
========================= 9 passed, 1 warning in 1.39s =========================
```

Full unit suite (`cd src && python -m pytest tests/unit/ -v`):

```text
=========== 4132 passed, 15 skipped, 2 warnings in 118.91s (0:01:58) ===========
```

(all 15 skips are pre-existing `need --slow option to run` gates)

Repo harness (`bash src/tests/scripts/run_tests.bash`): exit 0 — `✓ All tests passed!`

`pre-commit run --all-files`: exit 0 — all hooks pass (Black, isort, Flake8 source+tests, MyPy, Bandit, markdownlint, doc links, shellcheck, yamllint).

D4 coverage proof — running only `TestRealInitLoggingSystem` with `--cov=cascade_correlation`:

```text
_init_logging_system body lines 631-658 executed by the D4 test: 8/28 -> [632, 635, 636, 637, 640, 656, 657, 658]
```

Those 8 lines are exactly the 8 executable statements of the body (641–654 are continuation lines of the single `LogConfig(...)` call starting at 640) — 100% of the real body executes under pytest while the autouse fast-logging fixture remains in place for the rest of the suite.

## Notes

- No test was deleted, disabled, or weakened; the previously-pinned failure-path semantics (`False` returns, empty-list returns, `ValueError` raises) are all still asserted, both by the untouched existing tests and by new companion tests.
- Test snapshots are written only to pytest `tmp_path`; nothing under the repo tree (`src/snapshots/` committed `.h5` artifacts untouched).
- One resource correction from execution: the live conda env is `JuniperCascor1` (per AGENTS.md; `JuniperCascor` is now `*-DEPRECATED`).

### DIFF
--- a/src/cascade_correlation/cascade_correlation.py
+++ b/src/cascade_correlation/cascade_correlation.py
@@ -4725,14 +4725,14 @@ class CascadeCorrelationNetwork:
         cls,
         snapshot_path: Union[str, pl.Path] = None,
         restore_multiprocessing: bool = True,
-    ) -> bool:
+    ) -> Union["CascadeCorrelationNetwork", bool]:
         """
-        Restore the network state from a snapshot file.
+        Restore a network from a snapshot file.
         Args:
             snapshot_path: Path to the snapshot file
             restore_multiprocessing: Whether to restore multiprocessing state
         Returns:
-            bool: Success status
+            The restored CascadeCorrelationNetwork on success; False on failure
         """
         logger = Logger
         try:
@@ -4752,10 +4752,8 @@ class CascadeCorrelationNetwork:
                 logger.error(f"CascadeCorrelationNetwork: restore_snapshot: Failed to load network from snapshot: {snapshot_path}")
                 return False
 
-            # Copy loaded network state into current instance
-            cls.__dict__.update(loaded_network.__dict__)
             logger.info(f"CascadeCorrelationNetwork: restore_snapshot: Restored snapshot from {snapshot_path}")
-            return True
+            return loaded_network
         except Exception as e:
             logger.error(f"CascadeCorrelationNetwork: restore_snapshot: Error restoring snapshot: {e}")
             import traceback
@@ -5377,12 +5375,6 @@ class CascadeCorrelationNetwork:
         accuracy = 0.0
 
         # Validate input tensors
-        if x is None or y is None:
-            self.logger.error("CascadeCorrelationNetwork: calculate_accuracy: Missing required tensors for accuracy calculation, using safe defaults.")
-            self.logger.debug(f"CascadeCorrelationNetwork: calculate_accuracy: input size: {self.input_size}, output size: {self.output_size}")
-            x = torch.empty(0, self.input_size)
-            y = torch.empty(0, self.output_size)
-            # raise ValueError("CascadeCorrelationNetwork: calculate_accuracy: Missing required tensors for accuracy calculation.")
         if not (isinstance(x, torch.Tensor) and isinstance(y, torch.Tensor)):
             self.logger.error(f"CascadeCorrelationNetwork: calculate_accuracy: Input and target tensors must be of type torch.Tensor. Input (x): {type(x)}, Target (y): {type(y)}")
             raise ValueError("CascadeCorrelationNetwork: calculate_accuracy: Input and target tensors must be of type torch.Tensor.")
--- a/src/snapshots/snapshot_utils.py
+++ b/src/snapshots/snapshot_utils.py
@@ -8,7 +8,7 @@ import datetime
 import os
 import shutil
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import h5py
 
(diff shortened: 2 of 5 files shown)


pcalnon:
CI note: the sole red check (Docker Build & Smoke Test → Quality Gate) is the pre-existing `main` breakage, not this diff. `main`'s own CI fails the same job at this branch's base commit `0a44938a` (and its parent) — the smoke test's *Verify Health Endpoint* step gets no JSON from `/v1/health`, i.e. the published-container bind-contract issue that open PR #390 ("heals main") fixes. First rerun also hit a transient Docker Hub 502 pulling `python:3.14-slim`. Everything else is green: pre-commit (3.12/3.13/3.14), unit tests + coverage, integration, security scans, docs, build. Suggest merging #390 first, then re-running this PR's docker job.

pcalnon (commented):
approved

PR #396, opened 2026-07-10
fix(lifecycle): create network on start from dataset dims + translate staged configs to juniper-data schema — PR-B1

pcalnon:
## Summary

**PR-B1 of the 2026-07-09 training-start fix plan** (defect 1 — no network-creation path; diagnosis: juniper-ml `notes/JUNIPER_2026-07-09_JUNIPER-ECOSYSTEM_TRAINING-START-FAILURE-DIAGNOSIS-AND-FIX-PLAN.md` §4.1, lands via juniper-ml#633). Two coupled fixes, both in the lifecycle manager:

### 1. Create-on-start

With `auto_start` now defaulting off (only the deploy demo profile re-enables it), nothing created a network before the first user-initiated start — every `POST /v1/training/start` on a fresh service 409'd on the pre-lock "No network created" guard, **before the staged/pending dataset was even consumed** (verified live). `start_training` now consumes the pending dataset first and, when no network exists, creates one sized from the actual training arrays (`input=X.shape[1]`, `output=y.shape[1] if 2-D else 1`) — exact `_auto_start_training` parity. A bare start with neither data nor a staged dataset still fails loud with "Training data not provided". `create_network`'s body moves to `_create_network_locked` so the non-reentrant lifecycle lock is held across consume→create→start (public behavior unchanged).

### 2. Staged-config dialect translation

`StageDatasetRequest` speaks canopy's dialect — `dataset_type` Literal `'spirals'`/`'moons'`, **total** `n_samples`, `rotations` — while juniper-data's registry keys are `'spiral'`/`'moon'` with per-arm/per-quadrant counts (`n_points_per_spiral`, `n_points_per_quadrant`) and `n_rotations`. `_reload_dataset` forwarded the canopy dialect verbatim, so every canopy-staged reload — the cold-swap **and** live-swap flows share this path — would die at juniper-data with `Unknown generator 'spirals'`. The seam was masked because the unit stubs called `_reload_dataset` with juniper-data names directly. `_translate_staged_config` now maps the dialect at the fetch boundary only; stored configs (pending/current/rollback) stay canopy-facing, and generic `params` keep winning on conflict (`setdefault`).

## Requirements

References JR-CAN-TRAIN-003 (external-cascor phases: backend sync / dataset adapters — this makes the canopy-facing staged-dataset and first-start contracts actually executable).

## Testing

- New: create-on-start via inline data and via pending staged dataset (network dims asserted from fetched arrays), single-column output-size inference, spirals/xor translation (`generator`/params asserted at the `create_dataset` boundary), generic-params precedence, canopy-dialect preservation in `_current_dataset_config`.
- Updated the two tests that had pinned the old contract (`test_start_training_without_network` 409, route-level inline-data 409).
- `tests/unit/api/` full suite green (exit 0); lifecycle suites 168 passed. Pre-commit clean (black/isort/flake8/mypy/bandit/async-audit).

Companions: canopy#437 (PR-A failure propagation, merged-independent), canopy PR-B2 (default-dataset staging for the trivial case) to follow, juniper-ml#634 (PR-C).

https://claude.ai/code/session_0192fCcLFN9EaZmstwK5Yjax

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -1422,37 +1422,47 @@ class TrainingLifecycleManager:
         Returns:
             Network info dictionary
         """
+        with self._lock:
+            return self._create_network_locked(**kwargs)
+
+    def _create_network_locked(self, **kwargs) -> Dict[str, Any]:
+        """``create_network`` body for callers that already hold ``self._lock``.
+
+        ``self._lock`` is a non-reentrant ``threading.Lock``, so ``start_training``'s
+        create-on-start path (training-start diagnosis 2026-07-09 PR-B) cannot call
+        the public ``create_network`` from inside its locked section — it calls this
+        instead. Behavior is identical to the public method.
+        """
         from cascade_correlation.cascade_correlation import CascadeCorrelationNetwork
         from cascade_correlation.cascade_correlation_config.cascade_correlation_config import CascadeCorrelationConfig
 
-        with self._lock:
-            if self.state_machine.is_started():
-                raise RuntimeError("Cannot create network while training is active")
+        if self.state_machine.is_started():
+            raise RuntimeError("Cannot create network while training is active")
 
-            self._params = kwargs.copy()
-            config = CascadeCorrelationConfig.create_simple_config(**kwargs)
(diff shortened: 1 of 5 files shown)


pcalnon (commented):
approved

PR #397, opened 2026-07-12
fix(api): C1 — snapshot create tolerates null description; save failures carry reason (not 404); history serialized from a consistent copy

pcalnon:
## Summary

Roadmap unit **C1** (I-3 upstream half) of the plan of record
`juniper-ml/notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md` (§4 I-3, §7 C1 row). Anchors verified against `2926245c` (origin/main at branch time). Sibling unit C2a (params apply reporting) is being implemented in parallel; this PR touches only the snapshot surfaces.

### 1. Explicit-null `description` accepted (the live 2026-07-11 incident)

`SnapshotCreateRequest.description: str = ""` 422'd on explicit `null` — canopy's route seam posts `{"description": null}` for a blank description, producing `{'type': 'string_type', 'loc': ['body','description'], 'input': None}` and the doubled "Failed to create snapshot" toast. The field is now `str | None = ""` with a `mode="before"` validator normalizing `null` → `""`, so omission, explicit null, and a plain string are all equivalent (`src/api/routes/snapshots.py`).

### 2. Failed saves carry their reason with the correct status

Pre-C1 chain: serializer swallowed ANY exception into `False` (`snapshot_serializer.py` `save_network` except-arm) → lifecycle `save_snapshot` returned `None` → route 404 **"No network available to snapshot"** — a disk/HDF5 failure masqueraded as a missing network.

Mechanism (typed exception):

- New stdlib-only `src/snapshots/snapshot_errors.py` defines `SnapshotSaveError` (kept out of the heavy h5py/numpy/torch serializer module so the route imports it at module scope for free).
- `CascadeHDF5Serializer.save_network` keeps its ERROR + stacktrace logging, then raises `SnapshotSaveError(f"{type(e).__name__}: {e}") from e`.
- `TrainingLifecycleManager.save_snapshot` propagates it (logging the filepath); a falsy serializer result also raises (defensive belt). `None` now strictly means **no network**.
- `POST /v1/snapshots` maps `SnapshotSaveError` → **500** with the reason in the detail; both no-network cases keep their **404**; the success path is byte-identical. 500 (not 409) because a write failure is a server-side fault, not a state conflict.

Bool-contract callers audited and unchanged in behavior: `cascade_correlation.save_to_hdf5` (method-level `except Exception` → `False`), `auto_snap_best` (`manager.py:1242-1247`), and the swap pre/post snaps (`manager.py:2474-2480`, `:2584-2590`) already wrap in `try/except Exception` and stay failure-tolerant.

### 3. Write-isolation hardening (shipped scope + documented remainder)

`_save_training_history` iterated the live `network.history` lists while the training thread appends per-epoch entries, and `save_snapshot` takes no lock. A manager-side lock cannot help: the appends happen inside `cascade_correlation.py`'s training loop, which shares no lock with the serializer (the manager's `_topology_lock` is only used by its own boundary/topology readers). Shipped: `_snapshot_history_view` takes a shallow point-in-time copy (top-level dict + each list value, a handful of GIL-atomic ops) up front, and the whole history writer serializes from that view — a mid-training save can no longer crash on concurrent list mutation. A source pin plus a threaded-append regression keep the writer on the view. `_save_weight_history` already `list()`-copies the containers it iterates.

**Documented remainder (explicitly out of this unit's scope):** (a) per-element deep consistency — unit-metadata dicts and swap events are appended fully built and at most attr-backfilled, so the shallow copy prevents crashes but does not freeze element internals; (b) `_save_network_objects_helper` still reads the live weight/topology tensors, so a snapshot taken mid-cascade-add may capture a torn architecture view — full isolation there needs a pause-or-deep-copy design at the manager level (candidate for a follow-up unit alongside C3's Path-B work).

## Testing

Env `JuniperCascor1`, all from `src/`:

- New `tests/unit/api/test_snapshot_create_error_paths.py` (14 tests): explicit-null / omission / string / no-body accepted with `""` normalization; failed save → 500 + reason (and explicitly not the no-network 404); both 404 cases pinned; end-to-end null-description create writing a real snapshot to `tmp_path`; real write-failure → 500; lifecycle contract (None only for no-network, raise with chained cause, success unchanged).
- New `tests/unit/test_snapshot_serializer_error_and_isolation.py` (8 tests): `SnapshotSaveError` typing with chained cause; success returns `True`; history-view copy contract (independent lists, all keys, empty/missing, non-list passthrough); source pin; 5 saves under a concurrent history-churn thread all succeed.
- Updated 5 tests that pinned the defective swallow-to-`False` contract to the typed-error contract (same code paths, corrected expectation): `test_snapshot_serializer.py::test_save_to_invalid_path`, `test_snapshot_serializer_coverage_deep.py::test_save_network_exception_handling`, `test_snapshot_weight_history.py` length-mismatch pair (now also assert the reason text), `test_lifecycle_manager_coverage_ext.py::test_save_snapshot_serializer_failure_raises`.
- All test snapshot writes go to pytest `tmp_path` (never `src/snapshots/`, which holds committed .h5 artifacts).

Results:

- `python -m pytest tests/unit/api/` → **1687 passed**
- Snapshot surface (15 files incl. both new) → **345+22 passed**
- Serialization/swap integration (`--integration`, `--slow`, `--golden` incl. the golden roundtrip against the committed .h5 artifacts) → **68+5+1 passed** (on-disk format unchanged)
- `bash src/tests/scripts/run_tests.bash` → **4077 passed / 0 failed / 0 errors / 10 skipped (slow-gated)** in 182 s
- `pre-commit run --all-files` → all hooks green (black, isort, flake8, mypy, bandit, shellcheck, markdownlint, doc links)

Note: `src/snapshots/snapshot_errors.py` required `git add -f` — `.gitignore`'s blanket `**/snapshots` rule (line 65) ignores new files there; the module is force-tracked exactly like the directory's existing .py sources.

## Requirements

References JR-CASCOR-API-* (snapshot management surface) — plan-driven fix; see plan §4 I-3 for the incident evidence chain.

https://claude.ai/code/session_01Pe5UkXcNAHTniWVTYh4GSN

### DIFF
--- a/src/api/lifecycle/manager.py
+++ b/src/api/lifecycle/manager.py
@@ -3851,10 +3851,18 @@ class TrainingLifecycleManager:
         slot. The no-collision path is byte-identical to the prior
         behaviour, so the ``snapshot_YYYYMMDDTHHMMSSZ`` format remains the
         common case for canopy / REST consumers that parse it.
+
+        C1 (I-3): ``None`` is returned only for the no-network case. A
+        failed write raises ``SnapshotSaveError`` carrying the serializer's
+        reason — pre-C1 both collapsed to ``None`` and the API route mapped
+        a disk/HDF5 failure to the same 404 as a missing network. The
+        failure-tolerant callers (``auto_snap_best``, the swap pre/post
+        snaps) already wrap this call in ``try/except Exception``.
         """
         if self.network is None:
             return None
 
+        from snapshots.snapshot_errors import SnapshotSaveError
         from snapshots.snapshot_serializer import CascadeHDF5Serializer
 
         serializer = CascadeHDF5Serializer()
@@ -3874,10 +3882,17 @@ class TrainingLifecycleManager:
             filepath = snapshots_dir / f"{snapshot_id}.h5"
             suffix += 1
 
-        success = serializer.save_network(self.network, filepath, include_training_state=True)
+        try:
+            success = serializer.save_network(self.network, filepath, include_training_state=True)
+        except SnapshotSaveError as exc:
+            self.logger.error(f"Failed to save snapshot to {filepath}: {exc}")
+            raise
         if not success:
+            # Defensive: ``save_network``'s contract is raise-on-failure, but a
+            # stubbed / legacy serializer returning falsy must not turn into a
+            # phantom "no network" 404 downstream (C1 / I-3).
             self.logger.error(f"Failed to save snapshot to {filepath}")
-            return None
+            raise SnapshotSaveError(f"Snapshot save to {filepath} failed (serializer returned {success!r})")
 
         self.logger.info(f"Snapshot saved: {snapshot_id}")
         return {
--- a/src/api/routes/snapshots.py
+++ b/src/api/routes/snapshots.py
@@ -5,9 +5,10 @@ import logging
 import re
 
 from fastapi import APIRouter, HTTPException, Request
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
 
 from api.models.common import success_response
+from snapshots.snapshot_errors import SnapshotSaveError
 
 logger = logging.getLogger("juniper_cascor.api.routes.snapshots")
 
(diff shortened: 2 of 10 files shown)


pcalnon (commented):
approved

PR #398, opened 2026-07-12
fix(api): C2a — parameter apply reports applied/skipped per key (no silent hasattr drops)

pcalnon:
## Summary

Executes roadmap unit **C2a** of the training-runtime-defects plan (juniper-ml [`notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md) §4 I-4 / §7 C2a, cross-cutting theme T3): the parameter-apply response now accounts for **every** requested key, killing the silent `hasattr` drop.

**Before**: `_apply_params_unlocked` (`src/api/lifecycle/manager.py`) silently dropped requested keys that failed `hasattr(self.network, k)` — and non-whitelisted keys — while returning 200 with the full params echo. Canopy's post-apply verification then read getattr-defaults and diverged: the latent applied-yet-error generator diagnosed under I-4.

**After**: the success return is the unchanged params echo plus two **additive** fields:

- `applied: [key, ...]` — keys that landed, in application order
- `skipped: [{"key": ..., "reason": ...}, ...]` — reasons:
  - `not-updatable` — key outside `updatable_keys`
  - `no-such-attribute` — whitelisted key the live network object lacks (the silent-drop case)
  - `null-value` — None nested/lifecycle value from an internal caller (REST/WS boundaries strip None via `exclude_none=True`)

`applied` and `skipped` exactly partition the requested key set (regression-pinned).

## Surfaces

- **REST** `PATCH /v1/training/params` — fields ride the `data` envelope (route mechanics unchanged; 404 no-model and 422 candidate-pool paths untouched).
- **WS** `set_params` — the same dict rides the success ack's free-form `result` field; ack schema unchanged (verified against the installed `juniper_cascor_protocol` `CommandResponseEnvelope`). Comment added at the `_execute_command` branch.

## Deliberately unchanged

- Pydantic request-model validation: a bound violation still rejects the **whole body** 422 atomically — reporting, not partial-apply-on-validation-failure (plan §4 I-4 "Options & trade-offs"; partial apply was explicitly deferred in validation).
- GAP-WS-28 atomic rollback: on a setter failure the exception still propagates after full revert (no report on failure by construction).
- `start_training`'s internal `_apply_params_unlocked` call discards the return — unaffected.
- Consumers verified compatible: canopy adapter (`apply_params` / `_verify_apply_roundtrip` iterate requested keys only), cascor-client (`update_params` is a raw pass-through). `FakeCascorClient` parity lands in roadmap unit CL2.
- No existing cascor test pinned the exact response key-set; tests were updated additively only (original assertions all retained).

## Testing (conda env `JuniperCascor1`)

| Suite | Result |
|---|---|
| `src/tests/unit/api/` (`python -m pytest tests/unit/api/`) | **1681 passed**, 0 failed (junitxml-verified; stdout summary truncation is the known pre-existing display quirk) |
| `bash src/tests/scripts/run_tests.bash` (full unit suite) | **4060 passed**, 0 failed, 10 skipped (pre-existing GPU/optional skips), 182.6s |
| `tests/integration/api/test_candidate_pool_invariants.py --integration` | **18 passed** |
| `pre-commit run --all-files` | all hooks passed (black, isort, flake8, mypy, bandit, markdownlint, doc-links, shellcheck, yamllint) |

New regression coverage:

- Route (`TestAppliedSkippedReporting`, `src/tests/unit/api/test_api_runtime_params.py`): all-valid keys → all in `applied` + `skipped == []` + full params echo intact; whitelisted-but-absent attribute (delattr'd live network) → `skipped(no-such-attribute)` with the sibling key applied under 200; empty body → both empty; one out-of-bound key → atomic 422 with the in-range sibling NOT landing.
- Manager (`TestUpdateParamsAtomicity` additions, `src/tests/unit/api/test_lifecycle_manager.py`): non-whitelisted key → `skipped(not-updatable)` (assertions added to the existing test, none removed); absent-attribute reporting never materializes the attribute; None nested value → `skipped(null-value)`; applied/skipped partition every requested key.
- WS (`test_set_params_ack_result_carries_applied_skipped`, `src/tests/unit/api/test_websocket_control.py`): success ack `result` carries `applied`/`skipped` alongside the params echo.

## Requirements

References JR-CASCOR-API-* (runtime parameter-update surface); this PR implements plan unit C2a (I-4 contract half). Sibling unit C2b (bound/surface coherence) and canopy N5 (render applied/skipped, toast detail) follow separately per the roadmap.

https://claude.ai/code/session_01Pe5UkXcNAHTniWVTYh4GSN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **C2a (I-4 / T3) — parameter apply reports `applied`/`skipped` per key; the silent `hasattr` drop is gone.**
+  `_apply_params_unlocked` (`src/api/lifecycle/manager.py`) silently dropped requested keys that failed `hasattr(self.network, k)` — and non-whitelisted keys — while returning 200 with the full params echo,
+  the latent generator behind canopy's applied-yet-error verification divergence (juniper-ml [`notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md) §4 I-4 / §7 C2a).
+  The success response now accounts for EVERY requested key with two additive fields alongside the unchanged params echo: `applied: [key, ...]` (keys that landed, in application order) and `skipped: [{"key", "reason"}, ...]`
+  with reasons `not-updatable` (outside the whitelist), `no-such-attribute` (whitelisted but absent on the live network object — the silent-drop case), and `null-value` (None nested/lifecycle value from an internal caller; the REST/WS boundaries strip None via `exclude_none=True`).
+  Carried additively through `PATCH /v1/training/params` (`data`) and the WS `set_params` ack (`result`) — no schema change; existing consumers (canopy adapter, cascor-client pass-through) are unaffected, and `FakeCascorClient` parity lands in roadmap unit CL2.
+  Pydantic request-model validation is untouched: a bound violation still rejects the whole body 422 atomically (deliberate — reporting, not partial apply), and the GAP-WS-28 atomic-rollback contract is unchanged.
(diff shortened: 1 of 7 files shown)


pcalnon (commented):
approved

PR #419, opened 2026-07-21
feat(metrics): C7 phase 1 — scalar evaluation metrics (f1/precision/recall/ROC-AUC) on the metrics surfaces

pcalnon:
## Summary

Roadmap unit **C7 phase 1** of the canopy training-runtime defects plan (juniper-ml [`notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md`](https://github.com/pcalnon/juniper-ml/blob/main/notes/JUNIPER_2026-07-11_JUNIPER-CANOPY_TRAINING-RUNTIME-DEFECTS-PLAN.md) §4-U U-4 / §7 C7 / §13). Closes the **compute half of U-4** for the scalar classification metrics — **F1, precision, recall, ROC-AUC** — and surfaces them additively on the existing cascor metrics surfaces. The **display half is canopy N9**; the **curve / explainability artifacts** (confusion matrix, SHAP, permutation / feature importance, PDP, calibration / ROC / PR / lift / gain curves) are an explicit **phase-2** follow-up and are deliberately NOT built here (a documented seam is left in the new module).

## Computation design

- **Module.** New dependency-free, torch-native `src/api/lifecycle/classification_metrics.py`. **No scikit-learn is added** — cascor already depends on `torch>=2.10` and nothing in the sklearn stack, so the metrics are implemented from first principles (consistent with the project's research philosophy).
- **Binary vs multi-class.** The decode mirrors the engine's own `_accuracy` (`cascade_correlation.py`): a single output column is thresholded at 0.5 (positive-class P/R/F1); two-or-more columns are argmax-decoded against a one-hot target.
- **Averaging (deliberate).** Multi-class precision/recall/F1 use **macro** by default — equal per-class weight, which surfaces minority-class collapse; the right posture for the Q2 multi-dataset continual-training goal where class balance varies run-to-run. `weighted` is also implemented and selectable (`average=`). Binary problems report the positive class (`average` reported as `"binary"`).
- **ROC-AUC.** Rank-based Mann-Whitney U identity with average ranks for ties — **exact**, matches `sklearn.metrics.roc_auc_score`. Binary uses the raw single-column score; multi-class is **one-vs-rest over softmax probabilities**, macro-averaged across the classes present on the batch.
- **Undefined-case handling (degrade gracefully).** Every metric returns `null` with a machine-readable reason instead of raising or reporting a misleading number: `single_class` (target has one distinct class — no contrast), `empty_batch`, `invalid_output` (NaN/Inf). A per-class zero denominator inside an otherwise-defined macro/weighted average contributes `0.0` for that class (the `zero_division=0` convention) and is NOT flagged as whole-metric-undefined.
- **Cadence & cost.** Computed in the manager's metrics drain (`_extract_and_record_metrics`) over the **evaluation split** — the validation/test tensors (`_val_x`/`_val_y`, sourced from the dataset's `X_test`/`y_test`) when present, else the training split — **once per completed training step** (initial output pass + one per growth iteration), **NOT per inner epoch**. Cost is a single `torch.no_grad()` forward pass over the eval split per step, run **outside `_metrics_lock`**, and **skipped entirely** on the frequent within-pass drains that add no history row. Negligible for the 2-D research datasets; bounded for large sets because it is step-cadenced. The scalars are attached to the **terminal** training-step row of each drain only (a single forward pass reflects current network state, not each backfilled row).

## Surfaces (all additive & nullable)

- **Metrics-history rows** (`TrainingMonitor.on_epoch_end`) gain flat `f1` / `precision` / `recall` / `roc_auc` keys — always present (stable schema), populated only on the terminal training-step row. These flow automatically to `GET /v1/metrics/history`, the WS **`metrics`** frames (built directly from the row dict), and the WS **`initial_metrics`** burst.
- **`GET /v1/metrics` snapshot** (`get_metrics`) gains the same flat fields plus a self-describing **`eval_metrics`** block: `enabled`, `average`, `split`, `n_samples`, `n_classes`, `undefined`.
- **`/v1/training/status`**: unchanged — per-epoch loss/accuracy do not appear there, so per the task the scalars are added only where loss/accuracy already surface.

## Protocol-package compatibility

**No `juniper-cascor-protocol` release is required.** Verified against the installed package (0.1.0): `MetricsEnvelope.data` / `StateEnvelope.data` are `dict[str, Any]`, `InitialMetricsData.metrics` is `list[Any]`, and `BaseEnvelope` is `ConfigDict(extra="allow")` — additive keys validate. `model_dump(exclude_none=True)` strips only top-level envelope fields (e.g. `seq`), not the nullable keys inside `data` (empirically confirmed). No follow-up owed.

## Config

Enabled by default. `JUNIPER_CASCOR_EVAL_METRICS_ENABLED=0`/`false` disables the computation — distinct from `JUNIPER_CASCOR_METRICS_ENABLED` (which gates the Prometheus endpoint). Documented in the AGENTS.md env table.

## Phase-2 seam

The curve/explainability artifacts belong beside the new module (same `(output, target)` inputs, same binary-vs-multiclass decode) but are out of scope for phase 1; the module is kept scalar-only with a documented note.

## Tests

- `src/tests/unit/api/test_classification_metrics.py` — metric correctness pinned to **hand-computed confusion matrices** (binary P/R/F1 = 2/3; multi-class macro & weighted exact) and a **hand-ranked ROC-AUC** (5/6); the `single_class` / `empty_batch` / `invalid_output` degradations; binary vs multi-class paths; tie handling; input non-mutation; contract (invalid-average / non-tensor raise).
- `src/tests/unit/api/test_eval_metrics_pipeline.py` — real-drain surface (scalars on the terminal row only; skipped when no new row); `/v1/metrics` snapshot flat fields + metadata; WS `metrics` frame carries the fields; REST `/v1/metrics` + `/v1/metrics/history` presence; the disable toggle; and **no-regression pins** for the existing loss/accuracy pipeline.

## Suite results

- **Full unit suite** (`bash src/tests/scripts/run_tests.bash`): **4191 tests, 0 failures, 0 errors, 10 skipped** (`--slow` only) — junit XML authoritative (exit 0). Core coverage 97%.
- **pre-commit** (all changed files): all hooks pass (black, isort, flake8 512/complexity-15, mypy, bandit, markdownlint, doc-link validation).

## Notes

- Commits are **unsigned** (`--no-gpg-sign`) per owner approval 2026-07-16.
- **Never merge** — owner merges. §13 tracker note follows in the plan doc (juniper-ml).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


pcalnon (commented):
approved

PR #486, opened 2026-08-08
feat(settings): Wave 3.1 — experiment YAML config layer (service:-block projection, YAML > env)

pcalnon:
## Summary

Wave 3.1 of the CLI experimentation plan (§14; design §5.1/§5.2/§5.6 in juniper-ml `notes/JUNIPER_2026-07-29_JUNIPER-ECOSYSTEM_CASCOR-RECURRENCE-CLI-TEST-VALIDATION-EXPERIMENTATION-PLAN.md`): the experiment YAML config layer for cascor, exactly per the ratified mechanism.

## What's included

- **`ExperimentYamlSettingsSource`** — a custom `PydanticBaseSettingsSource` that parses the experiment YAML once and projects **only its `service:` block** into `Settings`. The stock `YamlConfigSettingsSource` reads the file's *top-level* mapping, and this model is `extra="ignore"` — the stock source would silently no-op the entire layer (the §5.2 rationale, verbatim).
- **Ratified precedence** `CLI/init > YAML > env > .env > defaults` via `settings_customise_sources`, per the plan's snippet. The YAML source is inserted **only when `JUNIPER_CASCOR_CONFIG_FILE` is set** → the layer is inert for every existing env/compose deployment (risk R-4), and a projection log line records which keys the YAML supplied.
- **Fail-loud before boot** (§5.6 rules 1/2/6): unknown top-level blocks; `schema_version` gated (integer 1..1); unknown `service:` keys (silent dropping made impossible — mirroring the `TrainingParams` `extra="forbid"` precedent); and the launcher-owned infra keys `host`/`port`/`juniper_data_url` + `eval_metrics_enabled` (which belongs in `runtime:` as process env) all raise `ExperimentConfigError`.
- **`--config PATH`** on `server.py` and `main.py` (operator convenience): sets the env var **before** the first `lru_cache`'d settings load (§5.2's explicitly-called-out ordering constraint). The experiment stack itself threads the env var through the uvicorn-factory launch and stays the bind owner (§6.1/§6.2 — YAML can never override the launcher's port allocation).

## Tests

`src/tests/unit/api/test_experiment_yaml_settings.py` — the precedence arms (inert without the env var; YAML beats env; init-kwargs beat YAML; env still wins for unprojected fields; non-`service:` blocks ignored) and the full §5.6 rejection matrix (all four infra keys parametrized, unknown service key, unknown top-level block, missing/future `schema_version`, missing file, non-mapping YAML, invalid YAML). **41 passed** together with the existing settings suite; all pre-commit hooks pass.

## Wave boundaries

Reference YAMLs (`conf/experiments/`) are Wave 3.2; the recurrence mirror is Wave 3.3 (in flight); the direct-CLI `training:`/`dataset:` mapping is W-11 (Wave 3.6).

https://claude.ai/code/session_01Di3TxAstqvPX3qVBiNJrjH

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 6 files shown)


pcalnon (commented):
approved

PR #504, opened 2026-08-10
fix(api): F-P4-1 — honor SpiralParams + radius-10 scale in the training-start spiral fallback

pcalnon:
## Summary

Root-cause fix for **F-P4-1** (CLI experimentation P4 headline finding): the cascor SERVICE path terminated every spiral training run at ~epoch 2 with 0–1 hidden units and chance accuracy, at every budget including the 2000-epoch baseline — while the direct CLI trained spiral for minutes.

### Root cause

`_generate_spiral_data` — the in-process fallback materialized for `dataset.generator: "spiral"` on `POST /v1/training/start` — had two defects:

1. **Param infidelity**: it read only the legacy `n_per_spiral` key, silently ignoring the juniper-data `SpiralParams` names every caller sends (`n_points_per_spiral`, `n_rotations`, `noise`, `seed`) → 200 points instead of the configured 1000, noiseless, unseeded.
2. **Degenerate unit-radius scale**: coordinates were normalized by 1/(4π) (|x| ≤ 1). With the default candidate init (`randn × 0.1`), every tanh candidate operates in its **linear regime**, and after output-layer convergence the residual is least-squares-orthogonal to all linear functions of the inputs — so best-of-pool correlation pins at **~2.7e-4** (observed identically in the live E-A runs and the in-process repro), improvement is ~6e-7/epoch, candidates early-stop unimproved, and `grow_network` breaks `below_threshold` at iteration 1.

The identical spiral scaled to the classic radius 10 (the default of both `spiral_problem` and juniper-data's `spirals` generator) recruits **12 hidden units to 0.995 train accuracy** under the same TrainingParams — reproduction committed as `util/ad-hoc/f_p4_1_spiral_service_repro.py` (arm A: 0 units / 0.505 acc / `below_threshold`; ×4π scale: 12 units / 0.995 / `max_iterations`).

### Fix

`_generate_spiral_data` now honors `n_points_per_spiral` (legacy `n_per_spiral` still accepted), `n_rotations`, `noise`, `radius`, and `seed`, and generates at **radius 10** by default. Point-count and rotation defaults are unchanged. 8 new regression arms in `TestGenerateSpiralData` pin the param fidelity, the radius-10 envelope, seeded-noise reproducibility, and the legacy alias.

### Companion changes (separate PRs)

- juniper-ml driver: stage spiral through `POST /v1/training/dataset` like every other generator, so experiments train on the *configured* juniper-data dataset (radius-10, seeded, noisy) rather than this fallback.
- juniper-cascor issue (filed separately): API `candidate_patience` / `candidate_convergence_threshold` are applied to the network but never threaded into worker `CandidateUnit` construction (pool always runs module defaults 50 / 0.001).

## Testing

- `pytest tests/unit/api/test_training_route_coverage.py` — 33/33 pass (25 existing + 8 new).
- Existing integration lifecycle test (`test_spiral_data_generator`) unaffected: legacy key honored, sample-count assertion unchanged.
- pre-commit (black/isort/flake8/mypy/bandit/async-audit) green on changed files.

## Evidence

- juniper-ml `notes/JUNIPER_2026-08-09_JUNIPER-ECOSYSTEM_CLI-EXPERIMENTATION-P4-STUDIES-EVIDENCE.md` §4 (F-P4-1)
- E-A suite artifacts: `~/.local/state/juniper-experiments/suites/e-a-cascor-budget-sweep-20260809T085929Z/` (12 registries + manifests)

https://claude.ai/code/session_01VjfDTFQXfJU71nLa96PJQV

### DIFF
--- a/src/api/routes/training.py
+++ b/src/api/routes/training.py
@@ -352,21 +352,44 @@ async def cancel_swap_dataset_live(request: Request) -> dict:
 
 
 def _generate_spiral_data(params: dict):
-    """Generate spiral dataset for training."""
+    """Generate the in-process fallback spiral dataset for training.
+
+    F-P4-1 (CLI experimentation P4): this fallback previously read only the
+    legacy ``n_per_spiral`` key and emitted a unit-radius spiral (coordinates
+    normalized by 1/(4pi)) — silent infidelities against the juniper-data
+    ``SpiralParams`` schema callers actually send. The unit-radius scale is
+    degenerate for candidate training: with the default candidate init
+    (``randn * 0.1``) every tanh candidate operates in its linear regime,
+    where the output-layer-converged residual is orthogonal to the candidate
+    family, so best-of-pool correlation pins at ~2.7e-4 and ``grow_network``
+    terminates ``below_threshold`` with zero hidden units at chance accuracy.
+    The identical spiral at the classic radius-10 scale (``spiral_problem``
+    and juniper-data both default to 10.0) trains normally.
+
+    Honors the juniper-data SpiralParams names (``n_points_per_spiral``,
+    ``n_rotations``, ``noise``, ``radius``, ``seed``); the legacy
+    ``n_per_spiral`` key is still accepted. Defaults preserve the prior point
+    count (100/spiral) and rotation count (2.0) while fixing the scale.
+    """
     import numpy as np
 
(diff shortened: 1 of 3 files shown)


pcalnon (commented):
approved

PR #535, opened 2026-08-19
feat(rename_snapshots): add script to rename snapshot files with uniq…

pcalnon:
…ue UUIDs

Introduces a new Bash script that renames snapshot files in the specified directory by appending a unique UUID to each filename. The script navigates to the application directory, processes files matching a specific pattern, and moves them to a designated target directory while providing pre- and post-move checks for verification. This enhances file management for snapshot artifacts.


pcalnon (commented):
approved

pcalnon (commented):
approved

PR #538, opened 2026-08-19
chore(sequence-safety): drop the (Advisory) suffix from the required check

pcalnon:
Drops the `(Advisory)` suffix now that the screen is a REQUIRED status check (juniper-ml#1011, 2026-08-18). A required gate labelled advisory is misleading.

The job name **is** the required-context string, so this lands as part of a three-phase sequence: the old context was removed from the ruleset first (otherwise this very PR would be blocked by a context it no longer publishes), and the new context is added after merge. The screen is unenforced only for the window between those steps.

Refs pcalnon/juniper-ml#1011


pcalnon (commented):
approved

PR #541, opened 2026-08-20
fix(snapshots): D-B — a corrupt snapshot is 422, not 404

pcalnon:
## Summary

Implements **D-B**: a corrupt snapshot is no longer reported as `404 Not Found`.

Restore / retrain / resume / replay all used to raise the same `404 "Snapshot 'X' not found or failed to load"`, fusing two opposite operator situations — *pick a different snapshot* and *investigate data loss*. A snapshot that exists but cannot be read now returns **`422`** with a machine-readable code; `404` is reserved for one that genuinely is not there.

Implements the design of record ([juniper-ml#1193](https://github.com/pcalnon/juniper-ml/pull/1193)) as decided by the owner on all four §10 questions: converge `start_replay`, `422` + machine-readable reason, and fold in the `verify_saved_network` wart.

## The shape

| Layer | Change |
|---|---|
| `snapshots/snapshot_load_status.py` (new) | The shared vocabulary — `SNAPSHOT_OK` / `SNAPSHOT_ABSENT` / `SNAPSHOT_CORRUPT` plus a `SnapshotLoadResult` whose `__bool__` follows `ok`, so `if not loaded:` call sites read unchanged. |
| `snapshot_serializer.py` | `load_network_result()` — the load still goes through `load_network`, and only a `None` return triggers a cheap second pass (`_classify_load_failure`) to work out why. |
| `snapshot_serializer.py` | `_validate_format_detail()` returns *which* check failed; `_validate_format()` is now a thin boolean wrapper over it. |
| `manager.py` | `_load_snapshot_to_network` returns the result; `_snapshot_result` carries `reason_code`; **`start_replay` converged** onto the same dict as its three siblings. |
| `routes/snapshots.py` | One `_raise_snapshot_load_error` helper behind all four verbs. |
| `app.py` | The error envelope accepts a dict detail carrying a semantic `error.code` — the API-09b path its own docstring already named. String details are untouched. |

## Two design decisions worth calling out

**`load_network` stays the seam.** My first cut had the manager call `load_network_result` and the serializer implement the classification inline. That broke 39 tests — not because the feature was wrong, but because I had moved the seam ~30 call sites patch. Inverting it (classify *after* a `None`, on the error path only) dropped the churn to 13, and every one of those 13 was a genuine contract change. The extra file open costs nothing on a path that only runs when a load has already failed.

**`409` was deliberately not reused.** These routes already spend `409` on FSM conflicts; reusing it for corrupt would trade one ambiguity for another. `500` was rejected too — it implies a retry might help, when a corrupt file is deterministic.

## `verify_saved_network`

`{"valid": False, "error": "Invalid format"}` came from the `_validate_format` gate *before any payload inspection* — so a snapshot with a perfectly good format string but a missing `params` group was reported as a format problem. It now names the failing check. A genuinely bad format still says "Invalid format", so the existing assertion at `test_snapshot_serializer_extended.py:78` keeps passing on its own merits.

## Verification

| Check | Result |
|---|---|
| New `tests/unit/api/test_snapshot_error_taxonomy.py` (25 tests) | pass |
| **Same file against the pre-D-B route mapping** | **12 fail** — all four verbs × both corrupt classes, plus the absent `code` arm |
| Full `tests/unit` (`--slow`) | pass, 0 failures |
| Full `tests/integration` (`--slow --integration`) | pass, 0 failures |
| **Golden trajectory** (`-m golden --golden`) | **pass** |
| `pre-commit` on all 10 changed files | pass |

The new suite covers each verb separately — the fusion was duplicated four times, and a fix applied to three is the likeliest way this ships half-done — plus two distinct corrupt classes (bad format vs. not-HDF5-at-all) so the classification is not keyed to one failure mode.

**The negative control matters most:** `test_valid_snapshot_still_succeeds` writes a real snapshot through the production serializer and asserts all four verbs return 200. Without it, a change that classified *everything* as corrupt would pass every other arm — and it correctly passes against both the old and new code, which is exactly what a control should do.

## Not in scope

`_validate_shapes` failure still only warns, and the shape-broken network is still installed on the live lifecycle. That is **D-E** ([juniper-ml#1199](https://github.com/pcalnon/juniper-ml/pull/1199)) — a different failure mode whose fix is a behaviour change over ~170 currently-loadable archive files, and it needs its own decision. A comment at the warn site now says so, so the next reader does not mistake it for an oversight. D-E's design notes that D-B had to land first, so that a rejected load has this `422` + reason channel to report through.

https://claude.ai/code/session_01741j36XEPuyk3Bh3nFrWJP

### DIFF
--- a/src/api/app.py
+++ b/src/api/app.py
@@ -765,12 +765,25 @@ def create_app(settings: Settings | None = None) -> FastAPI:
         """
         # Starlette's ``HTTPException.__init__`` auto-fills ``detail``
         # with ``HTTPStatus(status_code).phrase`` when the caller
-        # doesn't pass one, so ``exc.detail`` is always a string by
+        # doesn't pass one, so ``exc.detail`` is normally a string by
         # the time we get here. ``str()`` defends against future
         # subclasses that might return non-str detail objects.
-        message = str(exc.detail)
+        #
+        # API-09b (first use: D-B's snapshot error taxonomy): a route may instead pass
+        # ``detail={"code": "SEMANTIC_CODE", "message": "..."}`` when the caller needs a
+        # machine-readable discriminator that ``HTTP_NNN`` cannot express — two
+        # different failures sharing one status code, or one failure whose meaning
+        # clients must branch on without string-matching prose. String details are
+        # untouched, so every existing route keeps its exact shape.
+        detail = exc.detail
+        if isinstance(detail, dict) and "message" in detail:
+            code = str(detail.get("code") or f"HTTP_{exc.status_code}")
+            message = str(detail["message"])
+        else:
+            code = f"HTTP_{exc.status_code}"
+            message = str(detail)
         envelope = error_response(
-            code=f"HTTP_{exc.status_code}",
+            code=code,
             message=message,
         )
         return JSONResponse(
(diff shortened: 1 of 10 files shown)


pcalnon:
Superseded: re-opening with the sequence-safety waiver trailer in the single commit message. `_validate_format` shrank 85→11 lines because its body moved into the new `_validate_format_detail`, which the symbol-loss screen correctly flags as WEAKENED. The waiver has to live in one commit message so a squash-merge cannot drop it and turn main-verify red.

PR #544, opened 2026-08-20
delete(snapshots): remove obsolete snapshot files from the repository

pcalnon:
This commit deletes multiple outdated snapshot files from the `src/cascor_snapshots` directory. The removal of these files helps maintain a clean repository and prevents potential confusion with legacy data. All deletions have been verified to ensure no active references remain in the codebase.


pcalnon (commented):
approved

PR #546, opened 2026-08-20
fix(snapshots): one shared snapshot root at <repo>/cascor-snapshots

pcalnon:
## Summary

Implements the owner's 2026-08-20 snapshot storage ruling: **one snapshot root,
`<repo>/cascor-snapshots/`, shared by every stack origin** — the direct CLI, the FastAPI service, and
the container (which bind-mounts this same host directory, in the companion juniper-deploy PR).
Snapshots are project **assets**, not per-origin scratch: a model saved by a container run is
restored by a CLI run and resumed by a service run.

Supersedes two roots: `src/cascor_snapshots/` (historical direct-CLI) and `<repo>/snapshots/` (the
short-lived service root from #537, which never held a file). `JUNIPER_CASCOR_SNAPSHOTS_DIR` still
overrides both tiers, unchanged, blank-is-unset intact — that per-run override is how
`experiment_stack.bash` isolates experiments and stays the sanctioned opt-out.

Design of record: juniper-ml `notes/JUNIPER_2026-08-20_JUNIPER-ECOSYSTEM_SNAPSHOT-STORAGE-CONVENTION-DESIGN.md`.

## The load-bearing part is not what the first draft said

The first draft of this change claimed the **hyphen** in `cascor-snapshots` was a structural guarantee
— "not a valid Python identifier, so setuptools can never discover it as a package." **That is false**,
and three independent reviews caught it before this PR was opened. Verified directly:

```
$ python -c "from setuptools import find_namespace_packages; print([p for p in find_namespace_packages(where='.') if 'snapshot' in p])"
['cascor-snapshots', ...]
$ python -c "import importlib; print(importlib.import_module('cascor-snapshots'))"
<module 'cascor-snapshots' (namespace) ...>
```

`[tool.setuptools.packages.find]` in pyproject defaults to `namespaces = true` (PEP 420) — no
`__init__.py` requirement, no name filter beyond rejecting dots.

**So the real fix is `namespaces = false`**, which is now in `pyproject.toml`. That drops exactly
eleven directories that were being discovered as packages — `cascor-snapshots`, `conf`, `data`,
`docs`, `images`, `logs`, `notes`, `scripts`, `util`, **and the sibling distributions
`juniper-cascor-model` / `juniper-cascor-protocol`**, which publish their own wheels and must never
ship inside this one — while keeping all 29 real packages. Pinned by `TestPackagingExcludesArtifacts`.

The hyphen stays as defence in depth (it blocks the `import` statement form and plain
`find_packages`), just re-filed as taste rather than as the guarantee.

## Three latent breaks found and fixed on the way

None of these are caused by the rename; all three would have been carried forward by it.

1. **The systemd unit could not start.** `ReadWritePaths=` **fails** a unit when the path is missing,
   and two of its entries named directories that no longer exist. All entries now take the `-`
   prefix, and `cascor-snapshots/.gitkeep` is tracked so a fresh clone has the directory. That
   `.gitkeep` also fixes the docker case: a missing bind-mount source is created by the **daemon, as
   root**, which then EPERMs the uid-1000 container on every save — silently, since the save path only
   warns. The ignore rule is `/cascor-snapshots/*` + `!…/.gitkeep`, not the bare directory form,
   because git cannot re-include a path under an excluded directory.

2. **`snapshot_cli cleanup` was an unguarded glob-deleter.** It deleted immediately, no confirmation,
   `--keep 10` default. Pointed at the consolidated archive that removes **27,886 of 27,896** models.
   Worse, it sorted by `mtime`, which in this archive is *not* creation time (a copy reset every
   timestamp), so "keep the N most recent" did not select what it claimed to. Now dry-run by default
   (`--yes` to apply) and it refuses the shared root outright; `allow_shared_root=True` is reserved
   for a ratified retention policy. Verified: dry-run reports 5 and deletes 0; live deletes 5; the
   shared root with `keep=0, dry_run=False` returns 0 and leaves every file.

3. **The image had no correct default.** The service tier computed `parents[3]/"snapshots"` =
   `/app/snapshots` while compose mounted `/app/data`, one directory away — so every containerized
   snapshot went to the container's writable layer and died on recreate, with nothing erroring.
   `ENV JUNIPER_CASCOR_SNAPSHOTS_DIR=/app/cascor-snapshots` is now in the Dockerfile, so a bare
   `docker run` and the Helm path are right by default and orchestrators only mount over an
   already-correct path.

## The wheel no longer resolves a snapshot root inside the Python tree

`cascor_constants` is vendored verbatim into the published `juniper-cascor-model`, so this same file
runs from `site-packages` in the distributed worker. Repo-root-relative resolution would have made
that **worse**, not better:

```
BEFORE (src-relative)    <site-packages>/cascor_snapshots
NAIVE  (root-relative)   <python-lib>/cascor-snapshots      ← escapes site-packages entirely
```

Resolution is now: env override → repo root **when not imported from an install tree** → cwd with a
`RuntimeWarning` naming the variable. The install-tree test is `sysconfig` `purelib`/`platlib`
containment — stdlib-only, no subprocess, and exact. A `pyproject.toml`-adjacency probe was
considered and rejected: it answers *wrongly* inside the container, whose runtime stage copies only
`src/`. Both copies stay **byte-identical**, so `test_drift.py` needs no new
`_INTENTIONAL_DIVERGENCE` entry.

## Also

- `.dockerignore` excludes the artifact roots and `logs/` — nothing `COPY`s them, but every build
  walked ~5 GB of context, and a future `COPY . .` would have baked the archive into a published image.
- `conf/common.conf`'s third definition of the root (`SNAPSHOTS_DIR`) repointed — it is sourced by
  `util/script_template.bash` and `util/get_code_stats.bash`.
- `util/rename_snapshots.bash` `DEST_DIR`; snapshot-artifact docs in `docs/api/API_REFERENCE.md`,
  `docs/source/QUICK_START.md`, `docs/install/REFERENCE.md`, `notes/API_REFERENCE.md`.

## Testing

- `tests/unit/api/test_w6_snapshots_dir_override.py` — 21 tests. New: both tiers resolve to the
  **same** directory unset (the convention's actual guarantee), the three superseded roots are
  asserted against so a revert cannot pass silently, the installed-copy fallback warns and never
  lands under `purelib`/`platlib`, a configured deployment does **not** warn, and the five
  `TestPackagingExcludesArtifacts` arms pin `namespaces = false`.
- `juniper-cascor-model/tests/test_constants_dir_overrides.py` + `test_drift.py` — 15 tests, byte-identity intact.
- `tests/unit/api/` + snapshot serializer suites, `--slow`: exit 0.
- `pre-commit` clean on all 20 files. Note for reviewers: black reformats only the `src/` copy, so
  the vendored copy must be re-synced after formatting or the drift gate fails — done here.

## Requirements

Closes the design's **S-1** open question. Also closes **INT-P3-010** (`cascor_snapshots` vs
`snapshots` directory confusion).

https://claude.ai/code/session_016kVS1qAXn4ZTdCEoDUWwyR

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -32,3 +32,14 @@ dist/
 .ruff_cache/
 .coverage
 coverage.xml
+
+# Snapshot artifacts and logs. Nothing COPYs them (the Dockerfile takes only
+# requirements.lock, pyproject.toml/README/LICENSE, and src/), but every
+# `docker compose build` still walks and stats them as build context: 1.8 GB /
+# ~27,900 files of snapshots plus ~3.2 GB of logs, on a context that is otherwise
+# ~192 MB. This is also a correctness guard -- a future `COPY . .` would otherwise
+# bake the snapshot archive into a published image.
+cascor-snapshots/
+cascor_snapshots/
+snapshots/*.h5
+logs/
--- a/.gitignore
+++ b/.gitignore
@@ -50,30 +50,47 @@ __init__.cpython*
 
 # CascadeCorrelationNetwork snapshot ARTIFACTS.
 #
-# `src/snapshots/` is an importable Python package (snapshot_serializer.py,
-# snapshot_cli.py, snapshot_common.py, ...) that ALSO receives .h5 artifacts at
-# runtime. Six of the rules here used to ignore the whole directory
-# (`src/snapshots/*`, `src/snapshots/`, `src/snapshots`, and the `**/snapshots`
-# equivalents), which swept in that tracked source: editing any module in it
-# required `git add -f`, and a glob in this directory is a code-deletion tool --
-# a cleanup once removed five snapshot MODULES and broke every boot (cascor#501).
+# The ONE snapshot root is `<repo>/cascor-snapshots/` -- the repo root, shared by
+# every stack origin: the direct CLI, the FastAPI service, and the container,
+# which bind-mounts this same host directory. Snapshots are project ASSETS, so
+# they live inside the Juniper tree where the whole-tree offline backup captures
+# them, and they are ignored here so they never enter git. Transfer between hosts
+# or users is out-of-band and user-driven; nothing here is published.
 #
-# Ignore the artifacts by extension/name instead, so the package stays visible.
-**/cascor/cascor_snapshots/*
-cascor_snapshot_*.h5
+# DIRECTORY-anchored on purpose. The rules this replaces ignored by FILENAME
+# (`cascor_snapshot_*.h5`), so an artifact with any other name in the same
+# directory was NOT ignored -- `git check-ignore --no-index` proves it -- and the
+# service tier's own `snapshot_<ISO>Z.h5` never matched at all. Anchoring on the
+# directory means the filename can change (Phase 6.1 adds a run id) without
+# quietly un-ignoring 27,000 files.
+#
+# The `/*` form, not `/cascor-snapshots/`. Git cannot re-include anything under a
+# directory it has excluded, so the bare-directory form would make the .gitkeep
+# below impossible -- and that .gitkeep is load-bearing twice over: the systemd
+# unit's ReadWritePaths needs the directory to EXIST at start, and a missing
+# docker bind-mount source is created root-owned by the daemon, which then EPERMs
+# the uid-1000 container on every save. Shipping the directory is what makes a
+# fresh clone work. (`logs/*` at the top of this file uses the same form.)
+/cascor-snapshots/*
+!/cascor-snapshots/.gitkeep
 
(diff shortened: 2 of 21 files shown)


pcalnon:
Re-cutting: my `cleanup_old_files` dry-run guard broke four pre-existing test suites I did not run before opening (`test_snapshot_cli_coverage.py`, `test_snapshot_utils_extended.py`). Fixed, plus six new arms pinning the dry-run default, the shared-root refusal, its explicit opt-out, and that the guard does not block pruning a per-run RUN_DIR. Full `src/tests/unit` suite now run — exit 0 — and both sequence-safety screens reproduced locally before cutting.

